PR #30, opened 2019-05-29
fix GO111MODULE typo

icexin:
If the name of this environment variable is wrong, the compilation will degenerate to GOPATH mode.


duanbing:
LGTM

Issue #55, opened 2019-06-06
Add identity authentication process


MichelleZ:
**Brief of the issue**
Describe here
At present, there are two identities among XChain nodes: P2P node ID and XChain address. In order to solve the identity trust between nodes and prevent man-in-the-middle attack and message tampering, an authentication mechanism is needed between nodes, which can prove that the XChain address declared by symmetric nodes are true and effective.

**Your environment**
XuperUnion and OS/Go/gcc version etc.

**Repo steps**
Describe how to reproduce this issue.

**Expected behaviours**
Describe what you expected to happen.
Xchain address from peer is effective

**What actually happened**
Describe what actually happened. Include the full traceback if there was an exception.
Network is more secure

**Additional information**
If possible, include a minimal, complete, and verifiable example to help us identify the issue. This also helps check that the issue is not with your own code.

PR #56, opened 2019-06-06
Add identity authentication process

MichelleZ:
refer #55


core/xchainmg_net.go
```diff
@@ -390,3 +400,75 @@ func (xm *XChainMG) handleGetRPCPort(msg *xuper_p2p.XuperMessage) (*xuper_p2p.Xu
...
+		}
+
+		tsNow := time.Now().Unix()
+		tsPast, err := strconv.ParseInt(v.Timestamp, 10, 64)
+		if err != nil {
+			xm.Log.Error("handleGetAuthentication timestamp fmt error")
+			return errRes, errors.New("handleGetAuthentication timestamp fmt error")
+		}
+
+		if tsNow-tsPast >= config.DefautltAuthTimeout {
```
qizheng09: We usually use timers to make sure whether timeout. And the client always have a timer. Does the server need one?

core/xchainmg_net.go
```diff
@@ -390,3 +400,75 @@ func (xm *XChainMG) handleGetRPCPort(msg *xuper_p2p.XuperMessage) (*xuper_p2p.Xu
...
+			xm.Log.Error("handleGetAuthentication timestamp expired")
+			return errRes, errors.New("handleGetAuthentication timestamp expired")
+		}
+
+		data := hash.UsingSha256([]byte(v.PeerID + v.Addr + v.Timestamp))
```
yucaowang: v.PeerID should be the same with the remote peerid.

Issue #60, opened 2019-06-11
Add banning dapp contract


ToWorld:
**Brief of the issue**
Add banning dapp contract

**Your environment**
XuperUnion and OS/Go/gcc version etc.
OS: Linux
GO: 1.12.5
gcc: g++ (GCC) 4.8.2

**Expected behaviours**
reserved contract, when banning a dapp, the dapp can't be called any more in any way until the dapp is released again.

PR #63, opened 2019-06-11
add banning dapp contract

ToWorld:
issue #60


contractsdk/cpp/reserved/banned.cc
```diff
@@ -0,0 +1,83 @@
+#include "xchain/xchain.h"
+#include <vector>
+#include <string>
+
+class Banned : public xchain::Contract {};
+
+const char delimiter = ',';
+
+const std::vector<std::string> split(const std::string& rawStr) {
```
icexin: 这样返回vector会发生拷贝，作为参数传递进来会好些

ToWorld: Yes, It is correct. I have corrected it in the lastest version.

contractsdk/cpp/reserved/banned.cc
```diff
@@ -0,0 +1,85 @@
...
+        bool ret = ctx->delete_object(*iter);
+        if (!ret) {
+            ctx->error("release failed");
+            return;
+        }
+    }
+    ctx->ok("release contract success");
+}
+
+DEFINE_METHOD(Banned, get) {
```
qizheng09: This may need to support contract name list.

Issue #65, opened 2019-06-11
Reserved contracts mechanism.


qizheng09:
**Brief of the issue**
The reserved contracts mechanism for XuperUnion.

**Your environment**
OS: Linux
GO: 1.12.5
gcc: g++ (GCC) 4.8.2

**Description**
We want to provide a generic reserved contract mechanism. It configured by genesis block configuration. It will be invoked during pre-execution and validation.

PR #70, opened 2019-06-12
Reserved contracts mechanism

qizheng09:
The reserved contracts mechanism for XuperUnion. It configured by genesis block configuration. It will be invoked during pre-execution and validation. #65


contract/vm.go
```diff
@@ -18,10 +18,30 @@ var (
...
+type ContextConfig struct {
+	XMCache      *xmodel.XMCache
+	Initiator    string
+	AuthRequire  []string
+	ContractName string
+	GasLimit     int64
+}
+
+// NewContextConfig return context config of a transaction
+func NewContextConfig(xmCache *xmodel.XMCache, initiator string, authRequire []string, contractName string, gasLimit int64) (*ContextConfig, error) {
```
icexin: ContextConfig as a pure data object, does not require a constructor, directly fills the field with the stuct initialization syntax

utxo/utxo.go
```diff
@@ -1296,37 +1334,77 @@ func getGasLimitFromTx(tx *pb.Transaction) (int64, error) {
...
+			uv.xlog.Error("verifyTxRWSets NewContext error", "err", err, "contractName", tmpReq.GetContractName())
+			if i < len(reservedRequests) && err.Error() == "leveldb: not found" {
+				continue
+			}
+			return false, err
+		}
+
+		_, err = ctx.Invoke(tmpReq.MethodName, tmpReq.Args)
+		if err != nil {
+			ctx.Release()
```
fxsjy: when err == nil , ctx should also be released.

contract/vm.go
qizheng09: I see, thanks  for your comments.

utxo/utxo.go
qizheng09: ok~

utxo/acl_valid_verify.go
```diff
@@ -85,23 +85,27 @@ func (uv *UtxoVM) verifyRWACLValid(tx *pb.Transaction) (bool, error) {
...
+		ok, contractErr := pm.CheckContractMethodPerm(tx.AuthRequire, tx.AuthRequireSigns, digestHash, contractName, methodName, uv.aclMgr)
+		if !ok {
+			uv.xlog.Warn("tx info ", "AuthRequire ", tx.AuthRequire, "AuthRequireSigns ", tx.AuthRequireSigns)
+		}
+		if contractErr != nil {
+			return ok, ErrRWAclNotEnough
+		}
 	}
-	return ok, nil
+	return true, nil
```
ToWorld: You should return the variable of ok instead of true because the function of CheckContractMethodPerm can return a pair of <false, nil>. More details in the file of permission/permission.go

Issue #64, opened 2019-06-11
Add ip limit


ToWorld:
**Brief of the issue**
Add the function to limit the amount of same ip, including server and client limitation.

**Your environment**
XuperUnion and OS/Go/gcc version etc.
OS: Linux
Go: 1.12.5
gcc:  g++ (GCC) 4.8.2

**Expected behaviours**
Describe what you expected to happen.
Limiting the amount of same ip when a new stream generated, default the limitation of same ip is 10, and you can adjust the parameter in the config file.

yucaowang:
Close this issue as fixed #71

PR #71, opened 2019-06-12
add ip limit

ToWorld:
#64


p2pv2/stream_limit.go
```diff
@@ -0,0 +1,138 @@
...
+
+	if lg == nil {
+		lg = log.New("module", "p2pv2")
+		lg.SetHandler(log.StreamHandler(os.Stderr, log.LogfmtFormat()))
+	}
+	sl.log = lg
+}
+
+func (sl *StreamLimit) hasStream(addrStr string) bool {
+	if sl == nil {
```
qizheng09: There's no need to check whether `sl` is nil.

p2pv2/stream_limit.go
```diff
@@ -0,0 +1,138 @@
...
+		return false
+	}
+	_, exist := sl.streams.Get(addrStr)
+	return exist
+}
+
+// AddStream used to add the amount of same ip, plus one per call
+func (sl *StreamLimit) AddStream(addrStr string, peerID peer.ID) bool {
+	if sl == nil {
```
qizheng09: same as above

qizheng09:
LGTM

p2pv2/stream_limit.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/patrickmn/go-cache"
+
+	"github.com/xuperchain/log15"
+)
+
+// StreamLimit limit the peerID amount of same ip
+type StreamLimit struct {
+	// Store all streams available
+	// key: addr value: peerID
+	streams *cache.Cache
```
fxsjy: why use cache, not map ?

ToWorld: You can use map, sync.Map or cache.Cache, No much difference.

ToWorld: > why use cache, not map ?

I have corrected the issue using sync.Map instead of cache.Cache.

PR #73, opened 2019-06-13
candidate nominated should be auth

ToWorld:


consensus/tdpos/auth_address.go
```diff
@@ -0,0 +1,13 @@
+package tdpos
+
+func (tp *TDpos) isAuthAddress(candidate string, initiator string, authRequire []string) bool {
```
qizheng09: Please move  this function to `tdpos/common.go`

consensus/tdpos/run.go
```diff
@@ -171,6 +171,14 @@ func (tp *TDpos) runNominateCandidate(desc *contract.TxDesc, block *pb.InternalB
 	// 从内存中load出该候选人的记录
 	_, ok = tp.candidateBallots.Load(keyCanBal)
 	if !ok {
+		// check if the address nominated exists in the initiator or slice of auth_require
+		initiator := desc.Tx.GetInitiator()
+		authRequire := desc.Tx.GetAuthRequire()
+		if ok := tp.isAuthAddress(candidate, initiator, authRequire); !ok {
+			tp.log.Warn("candidate has not been authenticated", "candidate:", candidate)
+			return nil
```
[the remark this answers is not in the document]
ToWorld: > should return the error, why return nil ?

Yes, It is correct, I have corrected it.

consensus/tdpos/common.go
```diff
@@ -346,3 +346,15 @@ func (tp *TDpos) validateCheckValidater(desc *contract.TxDesc) (int64, int64, er
 	}
 	return version, term, nil
 }
+
+func (tp *TDpos) isAuthAddress(candidate string, initiator string, authRequire []string) bool {
+	if candidate == initiator {
+		return true
+	}
+	for _, value := range authRequire {
```
yucaowang: Note that initiator and authRequire could be Address or Account, but candidate should be address. So here should only check the "address part" of initiator and authRequire.

Issue #75, opened 2019-06-14
Add direct connections between core peers


yucaowang:
**Brief of your idea**
Add direct connections between core peers, so that core peers could send data directly to each other.

**What could be better if XuperChain using your idea**
1. Improve routing security, keep core network health.
2. Provide underly support for BFT-like consensus.

**Are there any side effects?**
No

**Additional information**
If possible, include a minimal, complete, and verifiable example to help us identify the issue.

yucaowang:
Close as fixed by #76

PR #76, opened 2019-06-14
Add direct connections between core peers.

yucaowang:
1. Add direct connections between core peers.
2. Add event bus for system events.
3. Support options for p2p message.

implements #75


common/events/eventbus.go
```diff
@@ -0,0 +1,171 @@
...
+func (ec *EventBus) Unsubscribe(eType EventType, handler EventHandler) error {
+	if handler == nil {
+		return ErrInvalidParams
+	}
+	ec.lock.Lock()
+	defer ec.lock.Unlock()
+	idx := ec.findHandlerIndex(eType, handler)
+	if idx < 0 {
+		// handler not found, treat as unsubscribe successfully
+		return ErrInvalidHandler
```
MichelleZ: Add HandlerNotExist = errors.New("handler not exist"),  this error description is more appropriate for the scene

common/events/eventbus.go
```diff
@@ -0,0 +1,171 @@
...
+
+// UnsubscribeMulti unsubscribe multiple events with given handler
+func (ec *EventBus) UnsubscribeMulti(eTypes []EventType, handler EventHandler) ([]EventType, error) {
+	if len(eTypes) == 0 || handler == nil {
+		return nil, ErrInvalidParams
+	}
+	failedTypes := []EventType{}
+	for _, eType := range eTypes {
+		err := ec.Unsubscribe(eType, handler)
+		if err != nil {
```
MichelleZ: err != nil && err != ErrHandlerNotExist

consensus/consensus_test.go
```diff
@@ -482,6 +482,11 @@ func TestUpdateConsensusOther(t *testing.T) {
 				"vote_unit_price":    "1",
 				"init_proposer": map[string]interface{}{
 					"1": []interface{}{"Y4TmpfV4pvhYT5W17J7TqHSLo6cqq23x3", "f3prTg9itaZY6m48wXXikXdcxiByW7zgk", "U9sKwFmgJVfzgWcfAG47dKn1kLQTqeZN3"},
+					"peerids": []interface{}{
```
MichelleZ: Should Peerids be included in round 1,
"1:", {addrs:{}, peerids:{}}

consensus/tdpos/tdpos.go
```diff
@@ -216,20 +217,33 @@ func (tp *TDpos) buildConfigs(xlog log.Logger, cfg *config.NodeConfig, consCfg m
...
 		return errors.New("TDpos init proposer length error")
 	}
 
+	// first round proposers' address
 	if _, ok := initProposer["1"]; !ok {
 		return errors.New("TDpos init proposer error, Proposer 0 not provided")
 	}
-	initProposer1 := initProposer["1"].([]interface{})
-	if int64(len(initProposer1)) != proposerNum {
-		return errors.New("TDpos init proposer error, Proposer 0 should be equal to proposerNum")
```
MichelleZ: If you agree with the above suggestion, you also need to make a simple change

core/xchaincore_handler.go
```diff
@@ -0,0 +1,90 @@
...
+		mcevent = msg.(*cons_base.MinersChangedEvent)
+	default:
+		xc.log.Warn("handleProposerChanged received unknown event message", "msg", msg)
+		return
+	}
+
+	cpi := &p2pv2.CorePeersInfo{
+		Name:           mcevent.BcName,
+		CurrentPeerIDs: make([]string, 0, len(mcevent.CurrentMiners)),
+		NextPeerIDs:    make([]string, 0, len(mcevent.NextMiners)),
```
MichelleZ: mcevent.CurrentMiners is a structure, is it filled in in another way?

common/events/eventbus.go
yucaowang: I think the return value should be nil in this case, just like the comments "handler not found, treat as unsubscribe successfully".

consensus/consensus_test.go
yucaowang: good point.

p2pv2/type.go
```diff
@@ -6,6 +6,66 @@ import (
 	p2pPb "github.com/xuperchain/xuperunion/p2pv2/pb"
 )
 
+// CorePeersInfo defines the peers' info for core nodes
+// By setting this info, we can keep some core peers always connected directly
+// It's useful for keeping DPoS key network security and for some BFT-like consensus
+type CorePeersInfo struct {
+	Name           string   // distinguished name of the node routing
+	CurrentTermNum int64    // the current term number
+	CurrentPeerIDs []string // current core peer IDs
```
qizheng09: Does p2p needs to know the term info?

consensus/base/types.go
```diff
@@ -0,0 +1,15 @@
...
+	Address  string // xchain address
+	PeerInfo string // peer info(in most cases is the network address)
+}
+
+// MinersChangedEvent define the information of miners would be changed.
+// this event would be fired when DPoS proposers initialized or next round proposers are selected.
+type MinersChangedEvent struct {
+	BcName        string
+	CurrentMiners []*MinerInfo
+	NextMiners    []*MinerInfo
```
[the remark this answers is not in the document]
yucaowang: Two terms info is not essential for current scenario, but is't useful to keep two term info for term  change and other filter strategies in the future(if possible).  E.g. we can implement a filter that get more peers from current term than next term.

p2pv2/type.go
yucaowang: It's better to know the current term number for implementing BFT-like consensus, but I still not found a good way to tell the accurate  term info to p2p. Currently no use.

p2pv2/server.go
```diff
@@ -92,19 +92,23 @@ func (p *P2PServerV2) Stop() {
...
 // 客户端再使用该方法请求带返回的消息时，最好带上log_id, 否则会导致收消息时收到不匹配的消息而影响后续的处理
-func (p *P2PServerV2) SendMessageWithResponse(ctx context.Context, msg *p2pPb.XuperMessage, fs FilterStrategy, withBreak bool) ([]*p2pPb.XuperMessage,
-	error) {
-	filter := p.getFilter(fs)
+func (p *P2PServerV2) SendMessageWithResponse(ctx context.Context, msg *p2pPb.XuperMessage,
+	opts ...MessageOption) ([]*p2pPb.XuperMessage, error) {
+	msgOpts := getMessageOption(opts)
+	filter := p.getFilter(msgOpts.filters)
 	peers, _ := filter.Filter()
+	withBreak := msgOpts.isBreak
```
qizheng09: It seems need to append msgOpts.targetPeerAddrs.

yucaowang: no need, targetPeerAddrs is reserved for the quick BFT consensus.

qizheng09: I see.

PR #79, opened 2019-06-14
ci pipeline

wenjingLi729:
ci pipeline script


test/gen_file.py
```diff
@@ -0,0 +1,44 @@
...
+#coding=utf-8
+import sys
+import json
+
+body1 = {
+    "module": "tdpos",
+    "method": "nominate_candidate",
+    "args": {
+        "candidate": "提名address"
+         }
```
MichelleZ: Direction connections between core peers will change this args part, this part'll contains addrs and peerID. So my suggestion is that this will be submited after that requirement is submitted.

Issue #80, opened 2019-06-14
Automatic rollback of transactions that have not been confirmed for too long


fxsjy:
**Brief of the issue**
In some cases, local node may have transactions that have been unsuccessfully confirmed, such as read/write set conflicts, and so on.Therefore, there needs to be a mechanism to clean up local unconfirmed transactions that have not been successfully confirmed for a long time.The default value is 300 seconds, which can be configured in the file of xchain.yaml.

**Your environment**
XuperUnion and OS/Go/gcc version etc.

**Repo steps**
make a network of tow nodes, one is miner, the other is verify node.
Then, shutdown the miner, and submit a transaction to verify node.
Wait 300 seconds, and then restart the miner.

**Expected behaviours**
The previous submitted transaction should be rollback, and we can see some undo logs in xchain.log.

yucaowang:
Close this issue as fixed #82

PR #82, opened 2019-06-14
Automatic undo transactions that have not been confirmed for too long

fxsjy:
#80


ToWorld:
> When the system rolls back transactions, it will roll back this more than 300 seconds, instead of triggering the rollback in 300 seconds. Is that right?

It marks unconfirmed transaction over 300 seconds and undoes the transaction, which is same as the flag of hasConflict, when doing the function of processUnconfirmTxs.

utxo/utxo.go
```diff
@@ -784,7 +788,11 @@ func (uv *UtxoVM) sortUnconfirmedTx() (map[string]*pb.Transaction, TxGraph, erro
 	var totalDelay int64
 	now := time.Now().UnixNano()
 	for txID, tx := range txMap {
-		totalDelay += (now - tx.Timestamp)
+		txDelay := (now - tx.Timestamp)
```
yucaowang: Just curious: If someone post a Tx with a fake timestamp which is ahead of current time for thousands of seconds, it the Tx will be pending in unconfirm Tx?  If so, how we can deal with such Tx using delay time?

fxsjy: yes, that is a bug. I have opened another issue: #92

yucaowang: I see

Issue #89, opened 2019-06-17
Reserved contract: identity contract.


qizheng09:
Brief of the issue
Add identity contract.

Description
A reserved contract which  be used for real-name authentication.

PR #90, opened 2019-06-17
add identity contract

qizheng09:
#89


contractsdk/cpp/reserved/identity.cc
```diff
@@ -0,0 +1,88 @@
...
+  const std::string initiator;
+  ctx->get_object(initiator, &value);
+  if (value != "true") {
+    ctx->error("verify initiator error");
+    return;
+  }
+
+  // FIXME zq: @icexin contex need to support auth_require
+  std::vector<std::string> auth_require;
+  for (auto iter = auth_require.begin(); iter != auth_require.end(); iter++) {
```
yucaowang: Could auth_require be forms like "account1/address1"? Which one should be identified as auth_require?

qizheng09: fixed

PR #91, opened 2019-06-17
add contract reserved character

ToWorld:
add contract reserved character


permission/acl/acl.go
```diff
@@ -45,3 +46,25 @@ func ValidRawAccount(accountName string) error {
...
+func validContractName(contractName string) error {
+	// param absence check
+	// contract naming rule check
+	contractSize := len(contractName)
+	contractMaxSize := utils.GetContractNameMaxSize()
+	contractMinSize := utils.GetContractNameMinSize()
+	if contractSize > contractMaxSize || contractSize < contractMinSize {
+		return fmt.Errorf("contract name length expect [%d~%d], actual: %d", contractMinSize, contractMaxSize, contractSize)
+	}
+	if m, _ := regexp.MatchString("^[a-z,A-Z,_]{1}[0-9,a-z,A-Z,_,.]{14}[0-9,a-z,A-Z,_]{1}$", contractName); !m {
```
icexin: It would be better to use a pre-compiled regexp. 
see https://godoc.org/regexp#MustCompile

permission/acl/acl.go
```diff
@@ -45,3 +46,25 @@ func ValidRawAccount(accountName string) error {
...
+		return fmt.Errorf("contract name does not fit the rule of contract name")
+	}
+	return nil
+}
+
+// ValidContractName check if contract name is ok
+// rules: contract name has the length of 16 including a-z,A-Z,0-9,_.
+// rules: the character of a-z,A-Z,_ is accepted.
+func ValidContractName(contractName string) error {
```
icexin: It's better to move this function to common package

yucaowang: +1

permission/acl/acl.go
```diff
@@ -45,3 +46,25 @@ func ValidRawAccount(accountName string) error {
...
+	}
+	if m, _ := regexp.MatchString("^[a-z,A-Z,_]{1}[0-9,a-z,A-Z,_,.]{14}[0-9,a-z,A-Z,_]{1}$", contractName); !m {
+		return fmt.Errorf("contract name does not fit the rule of contract name")
+	}
+	return nil
+}
+
+// ValidContractName check if contract name is ok
+// rules: contract name has the length of 16 including a-z,A-Z,0-9,_.
+// rules: the character of a-z,A-Z,_ is accepted.
```
yucaowang: two rules not the same?

permission/acl/acl.go
ToWorld: The package of permission/acl is a common package for permission verification.

permission/acl/acl.go
ToWorld: I have deleted the comments in the latest commit.

permission/acl/acl.go
ToWorld: > It would be better to use a pre-compiled regexp.
> see https://godoc.org/regexp#MustCompile

Thank you for you advice and I have fixed it.

permission/acl/acl.go
qizheng09: `ValidContractName ` is not the util function of permission. Its the common of xuperchain.

ToWorld: I have corrected it.

PR #95, opened 2019-06-17
 Optimizing connect to bootNodes failure code & Remove a superfluous line of code

af913337456:
优化了一下 DHT 连接 boot 节点失败判断条件的代码，避免了多次使用 len 读取一样的 cfg.BootNodes


fxsjy:
The CI build fail. It may be caused by gofmt diff. Could you please fix it? Thank you.

PR #108, opened 2019-06-19
Add global config for p2p core connection

yucaowang:
* add global config to enable/disable p2p core connection.
* use p2p stream pool to create core connections.
* update README.


yucaowang:
fix #108

PR #113, opened 2019-06-20
Add gas cost for memory and disk

icexin:
- add ResourceLimit for InvokeRequest in xchain.proto
- add gas cost for memory and disk
- update cpu cost table
- replace GasUsed to ResourceUsed in vm.Context
- replace GasLimit to ResourceLimit in vm.ContextConfig
- fix gas limit bug in wasm2c


cmd/cli/types.go
```diff
@@ -50,12 +50,18 @@ type TxOutputExt struct {
...
 type InvokeRequest struct {
-	ModuleName   string            `json:"moduleName"`
-	ContractName string            `json:"contractName"`
-	MethodName   string            `json:"methodName"`
-	Args         map[string]string `json:"args"`
+	ModuleName    string            `json:"moduleName"`
+	ContractName  string            `json:"contractName"`
+	MethodName    string            `json:"methodName"`
+	Args          map[string]string `json:"args"`
+	ResouceLimits []ResourceLimit   `json:"resource_limits"`
```
qizheng09: We now support multi-contracts calls. `ResouceLimits ` may need to share with other `InvokeRequest `.

contract/bridge/context.go
```diff
@@ -30,6 +31,23 @@ type Context struct {
 	Output *pb.Response
 }
 
+// DiskUsed returns the bytes written to xmodel
+func (c *Context) DiskUsed() int64 {
+	size := 0
+	_, wset, _ := c.Cache.GetRWSets()
+	for _, w := range wset {
+		size += len(w.GetKey())
+		size += len(w.GetValue())
```
yucaowang: is here any int overflow risks?

pb/xchain.proto
```diff
@@ -792,3 +793,15 @@ message IdentityAuth {
 message IdentityAuths {
     repeated IdentityAuth auth = 1;
 }
+
+enum ResourceType {
+    CPU = 0;
+    MEMORY = 1;
+    DISK = 2;
+    GAS = 3;
```
duanbing: GAS is too abstract as a ResourceType, cause resource can be converted to gas, the gas is one kind of Resource, looks confusing.  As a symbol of charge paid to the system contract,   equals to `Commission fees`  which we pay for the bank when we create a credit account.  so, I think we can use CFee instead of Gas will be more understandable.

cmd/cli/types.go
duanbing: one method for multiple contracts? @qizheng09

contract/bridge/context.go
icexin: fix later

pb/xchain.proto
icexin: XFEE my be a good name

cmd/cli/types.go
qizheng09: one method

PR #121, opened 2019-06-21
client tool to get sign from compliance check service

ToWorld:
Add a client tool to get sign from compliance check service


pb/xcheck.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+package pb;
+
+import "google/api/annotations.proto";
+import "xchain.proto";
+
+service Xcheck {
+    rpc ComplianceCheck(TxStatus) returns (ComplianceCheckResponse);
```
yucaowang: this is a client rpc proto, it's better keep the proto in cmd/cli.
pb folder is more like server-side proto.

cmd/cli/compliance_check.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return errors.New("Fail to open serialized transaction data file")
+	}
+	tx := &pb.Transaction{}
+	err = proto.Unmarshal(data, tx)
+	if err != nil {
+		return errors.New("Fail to Unmarshal proto")
+	}
+
+	tx.Txid, err = txhash.MakeTransactionID(tx)
```
yucaowang: Why here need a txid? at this time there is no signatures in txdata, so using MakeTransactionID cannot get the final txid.

pb/xcheck.proto
```diff
@@ -0,0 +1,15 @@
...
+}
+
+message ComplianceCheckResponse {
+    Header header = 1;
+	SignatureInfo signature = 2;
```
yucaowang: please format proto file using same indent characters.

ToWorld: Ok, I fix it.

cmd/cli/compliance_check.go
ToWorld: The client should pass TxStatus to the server.

pb/xcheck.proto
ToWorld: Yes, your advice is better and i will fix it.

cmd/cli/compliance_check.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	// XcheckClient
+	reply, err2 := c.XcheckClient().ComplianceCheck(ctx, txStatus)
+	if err2 != nil {
+		fmt.Println("check here new XcheckClient error", err2)
+		return err2
+	}
+	if reply.Header.Error != pb.XChainErrorEnum_SUCCESS {
+		return fmt.Errorf("Failed to post tx:%s, logid:%s", reply.Header.Error.String(), reply.Header.Logid)
+	}
```
MichelleZ: Post tx ? Fix it？

Issue #122, opened 2019-06-23
Set args for reserved contracts


ToWorld:
**Brief of your idea**
Set args for reserved contracts

**What could be better if XuperChain using your idea**
We need to add a function to set args for reserved contracts because each contract has its respective args.

**Are there any side effects?**
When a contract needs to set args differing from existing function, the code has to be updated.

qizheng09:
Use contract `context` to support it.

PR #123, opened 2019-06-23
add arg for reserved contracts

ToWorld:
#122


utxo/reserved_contract.go
```diff
@@ -35,12 +35,18 @@ func (uv *UtxoVM) getReservedContractRequests(req *pb.InvokeRPCRequest, tx *pb.T
...
+	// initialize contract name and method name
 	for _, v := range reservedContractCfgs {
-		req, err := parseReservedContractCfg(v)
+		reservedReq, err := parseReservedContractCfg(v)
 		if err != nil {
 			return nil, err
 		}
-		reservedContracts = append(reservedContracts, req)
+		// set args for all reserved contracts
+		reservedReq, err = setReservedContractArg(reservedReq, req)
```
icexin: Maybe setReservedContractArg should be called in parseReservedContractCfg?

PR #124, opened 2019-06-23
bug fix

ToWorld:
fix contract not found && remove GasNotEnough from getGasLimit instead of nil && fix banned contract return result


utxo/reserved_contract.go
```diff
@@ -13,12 +13,12 @@ func verifyReservedContractRequests(reservedReqs, txReqs []*pb.InvokeRequest) bo
...
 	for i := 0; i < len(reservedReqs); i++ {
-		if (reservedReqs[i].ModuleName != txReqs[i].ModuleName) || (reservedReqs[i].ContractName != txReqs[i].ContractName) ||
-			(reservedReqs[i].MethodName != txReqs[i].MethodName) {
+		if (reservedReqs[i].GetModuleName() != txReqs[i].GetModuleName()) || (reservedReqs[i].GetContractName() != txReqs[i].GetContractName()) ||
+			(reservedReqs[i].GetMethodName() != txReqs[i].GetMethodName()) {
 			return false
 		}
 		for k, v := range txReqs[i].Args {
-			if !bytes.Equal(reservedReqs[i].Args[k], v) {
+			if !bytes.Equal(reservedReqs[i].GetArgs()[k], v) {
```
ToWorld: Here, the function of GetArgs() is safer than Args.

utxo/utxo.go
```diff
@@ -1349,7 +1349,7 @@ func getGasLimitFromTx(tx *pb.Transaction) (int64, error) {
 		return gasLimit, nil
 	}
 	// FIXME: 没有小费的tx如何得到gas limit?
-	return 0, ErrGasNotEnough
+	return 0, nil
```
ToWorld: In the past, the function of getGasLimitFromTx was used to get gas limit for contract and It was nothing with the ordinary transfer transaction.
But, now we have to consider the return value of the function of getGasLimitFromTx has to be campatiable with all kinds of transactions because of reserved contracts, we change the return values of 0, nil instead of 0, ErrGasNotEnough.

utxo/utxo.go
```diff
@@ -1409,7 +1409,7 @@ func (uv *UtxoVM) verifyTxRWSets(tx *pb.Transaction) (bool, error) {
 		if err != nil {
 			// FIXME zq @icexin: need to return contract not found
 			uv.xlog.Error("verifyTxRWSets NewContext error", "err", err, "contractName", tmpReq.GetContractName())
-			if i < len(reservedRequests) && err.Error() == "leveldb: not found" {
+			if i < len(reservedRequests) && (err.Error() == "leveldb: not found" || strings.HasSuffix(err.Error(), "not found")) {
```
ToWorld: When the reserved contracts are not installed, It will return Key not found.

icexin: > When the reserved contracts are not installed, It will return Key not found.

Return an error which has an explicitly type will be better.

utxo/reserved_contract.go
icexin: > Here, the function of GetArgs() is safer than Args.

GetArgs ensure return a non nil slice, index error still exists.

utxo/utxo.go
qizheng09: Yes, model cache will support it.

PR #128, opened 2019-06-24
Add text audit error type

MichelleZ:
Add text audit error type for compliance check


pb/xchain.proto
```diff
@@ -34,6 +34,7 @@ enum XChainErrorEnum {
     RWACL_INVALID_ERROR = 34;
     GAS_NOT_ENOUGH_ERROR = 35;
     TX_VERSION_INVALID_ERROR = 36;
+    TEXT_AUDIT_CHECK_NOT_APPROVED = 37;
```
yucaowang: only add an error code for text audit? could this error code be more generic for auditing?

MichelleZ: If tx fails to pass text audit check interface, it's this error. Others are already included in the current types.

Issue #138, opened 2019-06-25
xvm: open module failed undefined symbol: sqrt


icexin:
**Brief of the issue**
deploy go wasm contract failed, found `undefined symbol: sqrt` in log

**Your environment**
OS: ubuntu 18.04
GO: 1.12.6

**Repo steps**
Describe how to reproduce this issue.

**Expected behaviours**
Describe what you expected to happen.

**What actually happened**
Describe what actually happened. Include the full traceback if there was an exception.

**Additional information**
If possible, include a minimal, complete, and verifiable example to help us identify the issue. This also helps check that the issue is not with your own code.

yucaowang:
close as fixed #139

PR #139, opened 2019-06-25
xvm: fix sqrt symbol not found in code.so

icexin:
issue #138


xvm/compile/compile.go
```diff
@@ -85,8 +85,8 @@ func CompileNativeLibrary(cfg *Config, target, source string) error {
 		"-o"+target,
 		"-I.",
 		"-I"+tmpdir,
-		"-lm",
 		csource,
+		"-lm",
```
fxsjy: So you put the option "-lm" in the end ? why it works

duanbing: > So you put the option "-lm" in the end ? why it works

csource needs math library.

PR #140, opened 2019-06-25
add tx forbidden contract

ToWorld:
try to fix the issue of #154


fxsjy:
why no description

ToWorld:
> why no description

I will fix it.

utxo/utxo.go
```diff
@@ -1913,6 +1914,50 @@ func (uv *UtxoVM) queryContractMethodACLWithConfirmed(contractName string, metho
...
+	if err != nil {
+		return false, false, err
+	}
+	ctx.Release()
+	// inputs as []*xmodel_pb.VersionedData
+	inputs, _, _ := modelCache.GetRWSets()
+	versionData := &xmodel_pb.VersionedData{}
+	if len(inputs) == 2 {
+		versionData = inputs[1]
+	} else {
```
yucaowang: to avoid unnecessary "else" which may cause code nesting, it's better to check "!=" and return.

utxo/utxo.go
```diff
@@ -1913,6 +1914,50 @@ func (uv *UtxoVM) queryContractMethodACLWithConfirmed(contractName string, metho
...
+	contextConfig.ContractName = request.GetContractName()
+	ctx, err := vm.NewContext(contextConfig)
+	if err != nil {
+		return false, false, err
+	}
+	_, err = ctx.Invoke(request.GetMethodName(), request.GetArgs())
+	if err != nil {
+		return false, false, err
+	}
+	ctx.Release()
```
yucaowang: Is it memory safe to return before ctx.Release?

contractsdk/cpp/reserved/forbidden.cc
```diff
@@ -0,0 +1,47 @@
...
+    xchain::Context* ctx = self.context();
+    // check if txid has been forbidden
+    const std::string key = ctx->arg("txid");
+    std::string value;
+    bool ret = ctx->get_object(key, &value);
+    if (ret) {
+        ctx->ok("txid has been forbidden");
+        return;
+    }
+    ctx->error("txid has not been forbidden");
```
yucaowang: error() will return using error code 500, so we indicate "not found" as "internal error"?

utxo/utxo.go
ToWorld: > Is it memory safe to return before ctx.Release?

Thanks, I will fix it.

utxo/utxo.go
ToWorld: I will fix it.

contractsdk/cpp/reserved/forbidden.cc
ToWorld: Here,  "Key not found" is same as the message of "txid has not been forbidden" because of the contract of forbidden being blacklist.

PR #144, opened 2019-06-25
support badger storage, log perf, async perf, p2p broadcast perf, and so on.

### LINKED ISSUE #84: support badger storage, log perf, async perf, p2p broadcast perf, and so on
**Brief of the issue**
support badger storage, log perf, async perf, p2p broadcast perf, and so on

ToWorld:
fix #84


common/errors.go
```diff
@@ -26,3 +28,13 @@ type ServerError struct {
 func (err ServerError) Error() string {
 	return pb.XChainErrorEnum_name[int32(err.Errno)]
 }
+
+func NormalizedKVError(err error) error {
```
yucaowang: is this a KV specified function? shall we put it in kv package?

ToWorld: Here, NormalizedKVError is aimed to normalize errors such as leveldb and badger. In the future, we might normalize other types of errors, so I think it is a good way to put NormalizedKVError into the file of common/errors.go.

yucaowang: OK

PR #153, opened 2019-06-27
Add idea ignore file & Modify the initialization location of a local xlog object

af913337456:
Modify:
1. Considering that some developers use idea's Goland editor with . idea folder, in order not to upload these files to git, I added filtering rules to . gitignore.
2. For MakeUtxoVM function, the original code initializes the xlog object after calling xlog, which may lead to null pointer exception.


utxo/utxo.go
```diff
@@ -32,12 +31,13 @@ import (
...
 	"github.com/xuperchain/xuperunion/pb"
 	pm "github.com/xuperchain/xuperunion/permission"
-	acl "github.com/xuperchain/xuperunion/permission/acl"
+	"github.com/xuperchain/xuperunion/permission/acl"
 	acli "github.com/xuperchain/xuperunion/permission/acl/impl"
 	"github.com/xuperchain/xuperunion/pluginmgr"
 	"github.com/xuperchain/xuperunion/utxo/txhash"
 	"github.com/xuperchain/xuperunion/vat"
 	"github.com/xuperchain/xuperunion/xmodel"
+	"os"
```
yucaowang: please notice the import organization separating system packages and 3rd party packages into different parts. so do not place "os" package with 3rd party packages.

qizheng09: yep

af913337456: Editor intelligent guide cause this, I fix it

af913337456:
@yucaowang Understood, the language of this PR has been modified

PR #155, opened 2019-06-29
Bugfix: read-only variable [const char* response] is not assignable

duanbing:
## Description
undefined behavior will be occured in wasm VM.  

Fixes # (issue)

## Type of change
Bugfix.  make sure it fixed in v3.2. 

## Brief of your solution

## How Has This Been Tested?


contractsdk/cpp/xchain/syscall.cc
```diff
@@ -14,9 +15,10 @@ static bool syscall_raw(const std::string& method, const std::string& request,
     if (response_len <= 0) {
         return true;
     }
+    std::vector<char> buf(response_len);
```
icexin: Why do you need a vector as a buffer here?

duanbing: The expression `response->data()` returns a `const char*`,  changing a const char* is an undefined behaviors.  std::vector<char> or char *buf = new (std::nothrow) char[response_len + 1] all work here.

icexin: Maybe `&(*response)[0]` the address of first element in `response` will be a better choice.
https://stackoverflow.com/a/16502000

PR #157, opened 2019-07-01
dpos status API return more information

### LINKED ISSUE #156: dpos status API return more information
**Brief of your idea**
To support `wallet` application. I will add more informations of the `dpos status` API, such as `proposers` and `proposer_num`.

qizheng09:
fix #156


pb/xchain.proto
```diff
@@ -708,6 +708,8 @@ message DposStatus
   int64 term = 1;
   int64 block_num = 2;
   string proposer = 3;
+  int64 proposer_num = 4;
+  repeated string proposers = 5;
```
yucaowang: proposer vs. proposers,  these two names may cause a little confusion?

pb/xchain.proto
```diff
@@ -708,6 +708,8 @@ message DposStatus
   int64 term = 1;
```
yucaowang: seems using different code indent style with the code above.

PR #163, opened 2019-07-02
query all accounts containing the specific address

### LINKED ISSUE #162: query all accounts containing the specific address
**Brief of your idea**
query all accounts containing the specific address

ToWorld:
fix #162


contract/kernel/ak_2_account.go
```diff
@@ -0,0 +1,111 @@
...
+	case pb.PermissionRule_SIGN_AKSET:
+		return updateForAKSet(ctx, akSets, accountName, method)
+	default:
+		return errors.New("update ak to account reflection failed, permission model is not found")
+	}
+	return nil
+}
+
+func updateAK2AccountReflection(ctx *KContext, aclOldJSON []byte, aclNewJSON []byte, accountName string) error {
+	if err := update(ctx, aclOldJSON, accountName, "Del"); err != nil {
```
yucaowang: Code logic is very clear. Just a suggestion, can we get the diff between old ACL and new ACL firstly so that we can avoid delete and add the same key.

utxo/utxo.go
```diff
@@ -1945,6 +1946,28 @@ func (uv *UtxoVM) queryContractMethodACLWithConfirmed(contractName string, metho
 	return uv.aclMgr.GetContractMethodACLWithConfirmed(contractName, methodName)
 }
 
+func (uv *UtxoVM) queryAccountContainAK(address string) (*pb.AK2AccountResponse, error) {
```
yucaowang: Usually we do not care RPC params in UTXO or xchaincore, it's better to return []string of accounts instead of pb.AK2AccountResponse, and pack the return value in server.go.

PR #220, opened 2019-07-17
feature: make all child DAGs sort separately

ToWorld:
## Description

make all child DAGs sort separately

Fixes # 219

## Type of change

## Brief of your solution

First, make all isolated points left, an isolated point means that nobody depends on it and it depends on nobody.
Second, we don't deal with the point which depends on nobody.
Third, we combine all isolated points into other points which depend on some points.

## How Has This Been Tested?
Regression Test


ToWorld:
> your unit test doesn't check the final sequence generated,

You are right, and I will add it.

PR #224, opened 2019-07-18
XuperCDT: add table engine and basic test framework

duanbing:
## Description
New Feature for SmartContract SDK:
1.  Add the hive-like table; 
2  Add basic test framework for contract;

Fixes # (issue)

## Type of change

## Brief of your solution

## How Has This Been Tested?


contract/bridge/syscall_service.go
```diff
@@ -168,16 +172,21 @@ func (c *SyscallService) NewIterator(ctx context.Context, in *pb.IteratorRequest
 		return nil, fmt.Errorf("bad ctx id:%d", in.Header.Ctxid)
 	}
 
+	limit := in.Cap
+	if limit == 0 {
```
ToWorld: replice limit == 0 with limit <= 0 because of the type of in.Cap is int32

ToWorld: replice limit == 0 with limit <= 0 because of the type of in.Cap is int32

contractsdk/cpp/table/basic_iterator.cc
```diff
@@ -0,0 +1,61 @@
+#include "table/basic_iterator.h"
+#include "xchain/xchain.h"
+
+namespace xchain { namespace cdt {
+
+Iterator::Iterator(xchain::Context* ctx, const std::string start, const std::string limit, size_t cap) {
```
ToWorld: replace  const std::string start with  const std::string& start?

contractsdk/cpp/table/basic_iterator.cc
```diff
@@ -0,0 +1,61 @@
...
+    _cap = cap;
+    _start = start;
+    _limit = limit;
+
+    if (!load()) {
+        _it = _buf.size();
+    } else {
+        _it = 0;
+    }
+}
```
MichelleZ: No mater what value is returned by load(), _it is all set to 0. So why did you write it by this way? What's your aim?

contractsdk/cpp/xchain/basic_iterator.cc
```diff
@@ -0,0 +1,78 @@
...
+            _buf.pop_back();
+        } else {
+            _start = _limit;
+        }
+    }
+    return true;
+}
+
+bool Iterator::next() {
+    bool ret = end();
```
MichelleZ: if the value of ret equals true, it means that _it is out of bounds. So next statement that getting the value belonged to index _it, is it illegal?

contractsdk/cpp/table/table.h
```diff
@@ -0,0 +1,90 @@
...
+class TableIterator;
+
+typedef std::pair<std::string, std::string> PairType;
+
+
+// Table implements table storage, with pb::Message as a record, supporting rowkey which should be unique,
+// and multiple indeies.
+// Layout on KV:
+//      M{table name}            -> max index id                           (1)
+//      K{rowkey}                -> row                                    (2)
```
MichelleZ: Should it be changed to {table_name}K{rowkey} here?

PR #232, opened 2019-07-25
quick shell to transfer & createAccount & deployContract & callContract

### LINKED ISSUE #233: Quick shell to play xuperchain
**Brief of your idea**
Use shell command to have a quick start with xuperchain, you just need to transmit necessary parameters.

It is more easier to have a start with xuperchain.

ToWorld:
## Description

quick start to play xuperchain

Fixes #233

## Type of change

## Brief of your solution
Wrap all necessary command in the shell file, client only needs to give paramters.

## How Has This Been Tested?

None


cmd/quick_shell/testnet.conf
```diff
@@ -0,0 +1,2 @@
+ip_port=""
+public_key=""
```
yucaowang: is public_key referring to compliance service's address?

cmd/quick_shell/deployContract.sh
```diff
@@ -0,0 +1,27 @@
...
+if [ ! -d "$acl" ]; then
+    mkdir $acl
+fi
+addrs="./data/acl/addrs"
+if [ ! -f "$addrs" ]; then
+    touch $addrs
+fi
+echo $public_key > data/acl/addrs
+echo "$accountName/$address" >> data/acl/addrs
+./xchain-cli wasm deploy --account $accountName --cname $contractName -H $ip_port -m $contractNamePath --arg '{"creator":"xchain"}' --output contractRawTx.out --fee $fee
```
yucaowang: "arg" should be a custom field

cmd/quick_shell/testnet.conf
ToWorld: Yes.

cmd/quick_shell/deployContract.sh
ToWorld: You are correct.

PR #265, opened 2019-08-02
Improvement: no need to forbid any tx if reserved contract is not available

### LINKED ISSUE #264: Improvement: no need to forbid any tx if reserved contract is not available
**Brief of your idea**
The contract of forbidden is available regardless of reserved contracts. It is not necessary to forbid any transaction when the blockchain system isn't under reserved contracts.

When the blockchain system is under non-reserved contracts, the contract of forbid doesn't work.

None

ToWorld:
## Description

When the xuperchain is under non-reserved contracts, the contract of forbidden couldn't work.

Fixes #264 

## Type of change

## Brief of your solution

Check if the rootConfig has the configurations of reserved contracts before invoking the contract of forbidden.

## How Has This Been Tested?

Regresstion Test

### DIFF
--- a/data/config/reservedroot.json
+++ b/data/config/reservedroot.json
@@ -52,14 +52,12 @@
                 "args":{}
             }
         ]
-        , "forbidden_contract": [
-            {
+        , "forbidden_contract": {
                 "module_name": "wasm",
                 "contract_name": "forbidden",
                 "method_name": "get",
                 "args":{}
-            }
-        ]
+        }
         , "permission": {
                 "CreateAccount" : { "rule" : "Null", "acl": {}},
                 "SetAccountAcl": { "rule" : "Null", "acl": {}},
--- a/data/config/reservedroot.json
+++ b/data/config/reservedroot.json
@@ -52,6 +52,14 @@
                 "args":{}
             }
         ]
+        , "forbidden_contract": [
+            {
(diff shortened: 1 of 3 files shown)


utxo/utxo.go
```diff
@@ -1710,10 +1710,19 @@ func (uv *UtxoVM) queryAccountContainAK(address string) ([]string, error) {
 }
 
 func (uv *UtxoVM) queryTxFromForbiddenWithConfirmed(txid []byte) (bool, bool, error) {
+	// 如果配置文件配置了封禁tx监管合约，那么继续下面的执行。否则，直接返回.
+	forbiddenContract := uv.ledger.GenesisBlock.GetConfig().ForbiddenContract
+	if len(forbiddenContract) != 1 {
```
qizheng09: This requires `len(forbiddenContract) ` equal 1, but the config of `xuper.json.forbidden_contract` is an array. emm, I not really understand.

ToWorld: In fact, I wanted to reuse the code. And I have corrected it.

data/config/reservedroot.json
```diff
@@ -52,6 +52,14 @@
                 "args":{}
             }
         ]
+        , "forbidden_contract": [
```
qizheng09: This also needs to be changed~

PR #272, opened 2019-08-05
Reserved contract support white list

### LINKED ISSUE #271: Reserved contract support white list
**Brief of your idea**
At present, all transactions need to be checked by  `ReservedContracts`. It is not suitable for some special scenarios.

The config of chain may support `ReservedWhitelist`, so that some transactions can not pass the inspection of `ReservedContracts`.

Nothing serious.

**Additional information**
Nothing.

qizheng09:
## Description

At present, all transactions need to be checked by ReservedContracts. It is not suitable for some special scenarios.

Fixes #271 

## Type of change

## Brief of your solution

The config of chain may support ReservedWhitelist, so that some transactions can not pass the inspection of ReservedContracts.
```
 ,"reserved_whitelist": {
     "account": "XC9999999999999999@xuper"
} 
```

## How Has This Been Tested?
It can be tested by changing `cmd/cli/cli.go` or calling `posttx` interface.


utxo/reserved_contract.go
```diff
@@ -23,7 +23,78 @@ func genArgs(req []*pb.InvokeRequest) *reservedArgs {
...
+		uv.xlog.Info("verifyReservedWhitelist true the utxo list is nil")
+		return true
+	}
+	fromAddr := string(tx.GetTxInputs()[0].GetFromAddr())
+	for _, v := range tx.GetTxInputs() {
+		if string(v.GetFromAddr()) != fromAddr {
+			uv.xlog.Info("verifyReservedWhitelist false fromAddr should no more than one")
+			return false
+		}
+		fromAddr = string(v.GetFromAddr())
```
yucaowang: is this line needed? fromAddr should be the same address for all TxInputs.

utxo/reserved_contract.go
```diff
@@ -23,7 +23,78 @@ func genArgs(req []*pb.InvokeRequest) *reservedArgs {
...
+		toAddrs[string(v.GetToAddr())] = true
+		if len(toAddrs) > 2 {
+			uv.xlog.Info("verifyReservedWhitelist false toAddrs should no more than two")
+			return false
+		}
+	}
+
+	// verify utxo output whitelist
+	for k := range toAddrs {
+		if acl.GetAksWeight()[k] <= 0 {
```
yucaowang: Two questions:
1. is it acceptable if the ak weight == 0 in ACL whitelist?
2. FromAddr may be one of the output ToAddr since UTXO split, is this check could cover the case?

PR #280, opened 2019-08-08
Dag parallel verify2

### LINKED ISSUE #261: parallel verification for the verification node
**Brief of your idea**
verifying the transactions of a block for the verification nodes receiving from master node concurrently and severally before doInternal makes it faster to synchronize blocks.

it will be faster for verification nodes to synchronize block.

ToWorld:
## Description

It is a kind of improvement of verify process from a lower verification process to a faster verification process.
Fixes #261 

## Type of change

- [x] It is a kind of improvement

## Brief of your solution

verifying the transactions of a block for the verification nodes receiving from master node concurrently and severally before doInternal makes it faster to synchronize blocks.

## How Has This Been Tested?

None


utxo/split_dag.go
```diff
@@ -0,0 +1,67 @@
...
+				continue
+			}
+			txGraph[refTxid] = append(txGraph[refTxid], txid)
+		}
+	}
+	return txMap, txGraph, nil
+}
+
+// SplitChildDAGs split child dags
+func SplitChildDAGs(txGraph TxGraph, outputTxList []string) []int {
```
fxsjy: I suggest migrate this logic of splitting into TopSort, so that can cost lesss computing.

PR #288, opened 2019-08-12
add ImmediateVerifyTx when Walking

### LINKED ISSUE #287: Walk function without enough verifytx cases to something error.
**Brief of the issue**
Walk function without enough verifytx cases to songmething error.

**Your environment**

**Repo steps**
step1
The bin of contract is not compatible between  machine A and machine B.
step2
Deploy cotract that some nodes can not execute successfully because of compatibility issues.
step3
But, the nodes that deploy contract failed can syncblock and have equal utxototal.

**Expected behaviours**

**What actually happened**

**Additional information**

ToWorld:
## Description

Add necessary verification when Walking in avoid to synchronize invalid blocks affecting the consistence of UTXO state.

Fixes #287

## Type of change

## Brief of your solution

Add necessary verification before doTx when Walking.

## How Has This Been Tested?

Regression Test.


utxo/utxo.go
```diff
@@ -1614,6 +1614,12 @@ func (uv *UtxoVM) Walk(blockid []byte) error {
 		idx, length := 0, len(todoBlk.Transactions)
 		for idx < length {
 			tx := todoBlk.Transactions[idx]
+			if !tx.Autogen && !tx.Coinbase {
+				if ok, err := uv.ImmediateVerifyTx(tx, false); !ok {
+					uv.xlog.Warn("dotx failed to ImmediateVerifyTx", "txid", fmt.Sprintf("%x", tx.Txid), "err", err)
+					return errors.New("dotx failed to ImmediateVerifyTx error")
```
fxsjy: How about just return the "err"?

ToWorld: From the point of my review, It is ok to throw out an abstract error instead of inner error.

PR #291, opened 2019-08-13
Miner was blocked when miner accepted an invalid block

### LINKED ISSUE #290: Other nodes are blocked when miner drug out an illegal node
**Brief of your idea**
The logic in doMiner() is that first of all, we need to do a loop to comprare utxo's lastedBlockID  with ledger's lastedBlockID，it will exit only if the two values are equal.
Other nodes are blocked when miner drug out an illegal node

We plan to add a switch, if this switch is true, we'll be continue to keep old logic; not, we'll remove this logic. So the process are not blocked.

Describe what might be worse.

**Additional information**

MichelleZ:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #290

## Type of change

## Brief of your solution
We plan to add a switch, if this switch is true, we'll be continue to keep old logic; not, we'll remove this logic. So the process are not blocked.

## How Has This Been Tested?


core/xchaincore.go
```diff
@@ -503,16 +506,18 @@ func (xc *XChainCore) doMiner() {
...
-	for !bytes.Equal(ledgerLastID, utxovmLastID) {
-		xc.log.Warn("ledger last blockid is not equal utxovm last id")
-		err := xc.Utxovm.Walk(ledgerLastID)
-		if err != nil {
-			xc.log.Error("Walk error ", "ledger blockid", global.F(ledgerLastID),
-				"utxo blockid", global.F(utxovmLastID))
-			return
+	if xc.walkLoop {
+		for !bytes.Equal(ledgerLastID, utxovmLastID) {
+			xc.log.Warn("ledger last blockid is not equal utxovm last id")
```
fxsjy: (1) Here,  we should call "Walk" at least  once.  But, your code will cause Walk not be called if the flag is false
(2) the option flag 'walkLoop' can be changed to another word, such as "failover". Or, "failfast" in opposite.

ledger/ledger.go
```diff
@@ -1027,3 +1027,29 @@ func (l *Ledger) MaxTxSizePerBlock() int {
...
+	}
+	newMeta.TrunkHeight = block.Height
+	metaBuf, pbErr := proto.Marshal(newMeta)
+	if pbErr != nil {
+		l.xlog.Warn("failed to marshal pb meta")
+		return pbErr
+	}
+	batchWrite.Put([]byte(pb.MetaTablePrefix), metaBuf)
+
+	l.meta = newMeta
```
fxsjy: No batchWrite.Write() ?

ledger/ledger.go
```diff
@@ -1027,3 +1027,34 @@ func (l *Ledger) MaxTxSizePerBlock() int {
 func (l *Ledger) GetBaseDB() kvdb.Database {
 	return l.baseDB
 }
+
+func (l *Ledger) Truncate(ledgerLastID []byte, utxovmLastID []byte) error {
```
icexin: Is ledgerLastID redundant?

MichelleZ: The purpose of retaining this parameter is  to deleting  incorrect blocks in the future  for next upgrade

Issue #294, opened 2019-08-15
Split utxo need preExec operation.


wenjingLi729:
**Brief of the issue**
Describe here
![image](https://user-images.githubusercontent.com/51076943/63072278-840bc500-bf55-11e9-83ea-bd43a0b69bc9.png)
![image](https://user-images.githubusercontent.com/51076943/63072309-b1587300-bf55-11e9-957c-d659ba1d3e8f.png)

**Your environment**
XuperUnion and OS/Go/gcc version etc.

**Repo steps**
Describe how to reproduce this issue.

**Expected behaviours**
Describe what you expected to happen.

**What actually happened**
Describe what actually happened. Include the full traceback if there was an exception.

**Additional information**
If possible, include a minimal, complete, and verifiable example to help us identify the issue. This also helps check that the issue is not with your own code.

ToWorld:
Closing this

PR #298, opened 2019-08-16
bugfix: split utxo need preexec and nominate address can be nil

### LINKED ISSUE #297: If the candidate is nil can not be nominated successfully.
**Brief of the issue**
Now,if the candidate is nil can be nominated successfully.
**Your environment**

**Repo steps**
step 1 
Change the file that candidate is nil.
step2
Nominate the candidate through the multisig function.
step3
query-candidates
image

**Expected behaviours**
If the candidate is nil can not be nominated successfully.

**What actually happened**

**Additional information**

qizheng09:
## Description

There are some problems.

- Split utxo cmd need to preExec before post to server.

- And the candidate can be nil while nominating.

Fixes #297 #294 

## Type of change

## Brief of your solution

- `Split utxo` cmd need to preExec before post to server.

- The candidate need to be address while nominating.

## How Has This Been Tested?


wenjingLi729:
if name == "" {
        return false
   }
There is return false,but the reference is judged to be true to enter execution.
if checkCandidateName(candidate) {}
There shpould be  judged as false.

PR #313, opened 2019-08-20
contract/native: add switch to diable native module

icexin:
Native contract module can be diabled via config file in mainnet.


common/config/config.go
```diff
@@ -143,6 +143,7 @@ type NativeConfig struct {
 	StopTimeout int
 	Deploy      NativeDeployConfig
 	Docker      NativeDockerConfig
+	Enable      bool
```
qizheng09: What if the node of the network is configured differently?

PR #335, opened 2019-08-26
optimization: use BlockHeader instead of Block with body

### LINKED ISSUE #336: optimization: use BlockHeader instead of Block with body
**Brief of your idea**
Speed up the process of syncForOnce

The miner can generate a new block on time

None

ToWorld:
## Description

Optimization: shorten the time of syncForOnce during the process of SyncBlocks to make sure that the miner generates a new block on time.

Fixes #336

## Type of change
- [x] Improvement 

## Brief of your solution

Use the function of QueryBlockWithHeader instead of the function of QueryBlock

## How Has This Been Tested?

Regression Test & Pressure Test

### DIFF
--- a/core/sync.go
+++ b/core/sync.go
@@ -46,7 +46,7 @@ func (xc *XChainCore) SyncBlocks() {
 	}
 }
 
-// syncForOnce sync block from peer nodes for one times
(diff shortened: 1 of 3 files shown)


ToWorld:
closing this

PR #337, opened 2019-08-27
add Pow interrupt

### LINKED ISSUE #338: Add pow interrupt mechanism
**Brief of your idea**
Add pow interrupt mechanism to avoid doing something uselss.

Decrease the possibility of forking.

The mechanism lacks of some strict check.

**Additional information**
None

ToWorld:
## Description

Add pow interrupt mechanism to avoid the frequent fork

Fixes #338 

## Type of change

- [x] Improvement

## Brief of your solution

1.Check the state regularly when minting in order to do something useless.
2.When get a new block ,check if it is valid and update the state.

## How Has This Been Tested?
Regression Test

### DIFF
--- a/core/valid_pow.go
+++ b/core/valid_pow.go
@@ -1,0 +1,36 @@
+package xchaincore
+
+import (
+	"github.com/xuperchain/xuperunion/common/log"
+	"github.com/xuperchain/xuperunion/consensus"
+	"github.com/xuperchain/xuperunion/ledger"
+	"github.com/xuperchain/xuperunion/pb"
+)
+
+func ValidPowBlock(block *pb.Block, xcore *XChainCore) bool {
+	internalBlock := block.GetBlock()
+	if xcore == nil || block == nil || internalBlock == nil {
+		log.Warn("invalid: xcore or block or internalBlock is nil")
+		return false
+	}
+
+	// validation for consensus of pow, if ok, tell the miner to stop mining
+	newBlockHeight := internalBlock.GetHeight()
+	if xcore.con.Type(newBlockHeight) == consensus.ConsensusTypePow {
+		if newBlockHeight < xcore.Ledger.GetMeta().GetTrunkHeight() {
+			log.Warn("invalid block: new block's height is not enough")
+			return false
+		}
+		actualTargetBits := internalBlock.GetTargetBits()
+		if !ledger.IsProofed(internalBlock.Blockid, actualTargetBits) {
+			log.Warn("receive a new block actual difficulty doesn't match blockid")
+			return false
+		}
+
+		//xcore.Ledger.StartPowBlockState()
+		// a valid new block shows up, let's interrupt the process of the miner to welcome it.
+		xcore.Ledger.AbortPowBlockState()
+	}
+
+	return true
+}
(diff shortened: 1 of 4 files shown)


core/valid_pow.go
```diff
@@ -0,0 +1,36 @@
...
+	if xcore == nil || block == nil || internalBlock == nil {
+		log.Warn("invalid: xcore or block or internalBlock is nil")
+		return false
+	}
+
+	// validation for consensus of pow, if ok, tell the miner to stop mining
+	newBlockHeight := internalBlock.GetHeight()
+	if xcore.con.Type(newBlockHeight) == consensus.ConsensusTypePow {
+		if newBlockHeight < xcore.Ledger.GetMeta().GetTrunkHeight() {
+			log.Warn("invalid block: new block's height is not enough")
```
fxsjy: it whould be better for trouble shooting, if blocks'height & trunk height are printed here.

core/valid_pow.go
```diff
@@ -0,0 +1,36 @@
...
+			return false
+		}
+		actualTargetBits := internalBlock.GetTargetBits()
+		if !ledger.IsProofed(internalBlock.Blockid, actualTargetBits) {
+			log.Warn("receive a new block actual difficulty doesn't match blockid")
```
fxsjy: print internalBlock.Blockid and proposer here

ledger/calc_nonce.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func (l *Ledger) processFormatBlockForPOW(block *pb.InternalBlock, targetBits int32) (*pb.InternalBlock, error) {
+	var gussNonce int32
+	var gussCount int64
+	valid := false
+	var err error
+
+	for {
+		if gussCount%65535 == 0 && !l.GetPowBlockState() {
```
fxsjy: magic number ..

ledger/ledger.go
```diff
@@ -1128,3 +1128,24 @@ func (l *Ledger) Truncate(utxovmLastID []byte) error {
...
+
+// StartPowBlockState 更新powBlockState状态
+func (l *Ledger) StartPowBlockState() {
+	l.powMutex.Lock()
+	defer l.powMutex.Unlock()
+	l.powBlockState = true
+}
+
+// AbortPowBlockState 中断挖矿流程
+func (l *Ledger) AbortPowBlockState() {
```
fxsjy: maybe, AbortPowMinning is better for understanding.

ledger/ledger.go
```diff
@@ -1128,3 +1128,24 @@ func (l *Ledger) Truncate(utxovmLastID []byte) error {
 	l.xlog.Info("truncate blockid succeed")
 	return nil
 }
+
+// StartPowBlockState 更新powBlockState状态
```
fxsjy: Comments for public method should be written in English.

ledger/ledger.go
ToWorld: Yes, all suggestions above are correct and I have updated them. Thank you.

ledger/ledger.go
ToWorld: I got it.

PR #366, opened 2019-09-08
Create account support proposal

### LINKED ISSUE #364: Proposal: NewAccountResourceAmount supports the proposal
**Brief of your idea**
It is not convenient for the current way storing the amount of creating a contract account, which doesn't change dynamically when it is necessary.
Therefore, We need to support a mechanism to change the amount of creating a contract account.

In different cases, you can choose different value for creating a new contract account if necessary.

None

ToWorld:
## Description

Update the amount of creating an new contract account according to the requirement dynamically.

Fixes #364 

## Type of change

## Brief of your solution

Use the mechanism of proposal to support the update of the amount of creating an account.
Pass the amount of creating an account into KContext.

## How Has This Been Tested?

Regression Test


icexin:
NewAccountResourceAmount does not need to be exposed to the contract sdk.

ToWorld:
Closing this pr and there is an another pr to replace it

PR #400, opened 2019-10-08
mv Consensus Parameters from Ledger to Utxo

ToWorld:
## Description

Consensus Parameters such as MaxBlockSize, ForbiddenContract and ReservedContracts was maintained in the ledger which is not atomic, causing the problem of not effectiveness of it.
It's a better way to maintain the consensus parameters in the utxo so that the system can prevent the question above.

Fixes # (issue)

## Type of change

## How Has This Been Tested?
Regression Test


contract/contract.go
```diff
@@ -21,6 +21,18 @@ var AutoGenWhiteList = map[string]bool{
 	"kernel.UpdateMaxBlockSize":  true,
 }
 
+// UtxoMetaRegister in avoid to being refered in a cycle way
+type UtxoMetaRegister interface {
```
yucaowang: Is it better to wrap all genesis configs in a struct, and make the struct one member of UTXOMeta?
By naming UtxoMetaRegister, it's more like a register for all UTXOMeta members, but it's just for genesis configs.

contract/kernel/kernel.go
```diff
@@ -200,6 +200,22 @@ func (k *Kernel) CreateBlockChain(name string, data []byte) error {
...
+	if err != nil {
+		k.log.Warn("GenerateRootTx error", "fullpath", fullpath)
+		os.RemoveAll(fullpath)
+		return err
+	}
+	txlist := []*pb.Transaction{tx}
+	k.log.Trace("Start to ConfirmBlock")
+	b, err := ledger.FormatRootBlock(txlist)
+	if err != nil {
+		k.log.Warn("format block error")
```
yucaowang: It's better to print the error message in Warning log since this err info is not returned.

contract/contract.go
ToWorld: Thank you for your suggestions. And I plan to give a reasonable name instead of UtxoMetaRegister and all the genesis configures have been the members of UtxoMeta.

PR #401, opened 2019-10-08
ignore 0 amount for output

ToWorld:
## Description

When the award amount is 0, it is a better way to ignore it when put utxo into leveldb which economizing some disk and speeding up the process of query.

Fixes # (issue)

## Type of change

- [ ] Improvement

## Brief of your solution

Ignore 0 amount for output

## How Has This Been Tested?

Regression Test

### DIFF
--- a/utxo/utxo.go
+++ b/utxo/utxo.go
@@ -1077,13 +1077,14 @@ func (uv *UtxoVM) undoTxInternal(tx *pb.Transaction, batch kvdb.Batch) error {
 		if bytes.Equal(addr, []byte(FeePlaceholder)) {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


utxo/utxo.go
```diff
@@ -1073,6 +1077,9 @@ func (uv *UtxoVM) undoTxInternal(tx *pb.Transaction, batch kvdb.Batch) error {
 		if bytes.Equal(addr, []byte(FeePlaceholder)) {
 			continue
 		}
+		if big.NewInt(0).SetBytes(txOutput.Amount).Cmp(big.NewInt(0)) == 0 {
```
fxsjy: line 1080 and line 1079 do the same convert,` big.NewInt(0).SetBytes(txOutput.Amount)`,

PR #402, opened 2019-10-09
compress the block to be broadcasted with snappy

ToWorld:
## Description

Make the size of block smaller than before when broadcasting in order to economize width.

Fixes # (issue)

## Type of change

- [x] Improvement

## Brief of your solution

Compress the block to be broadcasted with snappy

## How Has This Been Tested?

Regression Test


core/xchaincore.go
```diff
@@ -311,7 +312,8 @@ func (xc *XChainCore) repostOfflineTx() {
 		}
 		xc.log.Debug("repost batch tx list", "size", len(batchTxMsg.Txs))
 		msgInfo, _ := proto.Marshal(batchTxMsg)
-		msg, _ := xuper_p2p.NewXuperMessage(xuper_p2p.XuperMsgVersion1, xc.bcname, header.GetLogid(), xuper_p2p.XuperMessage_BATCHPOSTTX, msgInfo, xuper_p2p.XuperMessage_SUCCESS)
+		got := snappy.Encode(nil, msgInfo)
```
qizheng09: It's better to compress in `NewXuperMessage ` function so that the `p2p principle` can be migrated to other module easily.

p2pv2/pb/message.proto
```diff
@@ -62,6 +62,7 @@ message XuperMessage {
         MessageType type = 5;
         uint32 dataCheckSum = 6;
         ErrorType errorType = 7;
+		bool compressed = 8;
```
icexin: Using compress type is good for replacing other compression algorithms.

core/xchainmg_net.go
```diff
@@ -90,12 +90,9 @@ func (xm *XChainMG) handleReceivedMsg(msg *xuper_p2p.XuperMessage) {
 }
 
 func (xm *XChainMG) handlePostTx(msg *xuper_p2p.XuperMessage) {
-	txStatus := &pb.TxStatus{}
-
-	// Unmarshal msg
-	err := proto.Unmarshal(msg.GetData().GetMsgInfo(), txStatus)
-	if err != nil {
-		xm.Log.Error("handlePostTx Unmarshal msg to tx error", "logid", msg.GetHeader().GetLogid())
+	txStatus, err := handlePostTxMsgFromRemoteNode(msg)
```
yucaowang: It's better to wrap compress process in p2p library, you can leverage p2pv2.MessageOption to provide WithCompress option for SendMessage/SendMessageWithResponse.

PR #415, opened 2019-10-17
Performance Improvement: tx/block compression which saved bandwidth

### LINKED ISSUE #414: Performance: Compression of Transaction & Block
**Brief of your idea**
Speed up the transaction process per second by compressing the transaction and block.

Get higher tps than before.

None

ToWorld:
## Description

Speed up the transaction process amount per second.

Fixes #414 

## Type of change

## Brief of your solution

Compress transactions and blocks to be broadcasted with the algorithm of snappy.

## How Has This Been Tested?

Regression Test


p2pv2/pb/util.go
```diff
@@ -39,6 +41,35 @@ func CalDataCheckSum(msg *XuperMessage) uint32 {
 	return crc32.ChecksumIEEE(msg.GetData().GetMsgInfo())
 }
 
+// Compressed compress msg
+func Compressed(msg *XuperMessage) *XuperMessage {
```
yucaowang: Compress or Compressed?

ToWorld: > Compress or Compressed?

It's a better way to use compress. Thank you for your advice.

PR #416, opened 2019-10-17
xmodel: add lru cache for VersionedData

icexin:
xmodel.Get needs to get the tx data from the ledger to construct VersionedData, which generates a lot of proto.Unmarshal(tx) overhead.
This commit adds a lru cache for each bucket to xmodel to reduce overhead.

Benchmark env:

worker: 300
request_per_worker: 200
contract:counter

```
	old	new	delta
qps	1653	3267	+97.64%
```


icexin:
FlameGraph 

before:
image

after:
image

PR #427, opened 2019-10-22
support irreversible block from genesis json

### LINKED ISSUE #426: Function: support the function of irreversible block
**Brief of your idea**
Give a pair of variable <irreversibleBlockHeight, irreversibleSlideWindow> to record the checkpoint of irreversible place.
When calling the function of PlayAndRepost and PlayForMiner, try to update the variable of irreversibleBlockHeight.
When calling the function of runUpdateIrreversibleSlideWindow, try to update the variable of irreversibleSlideWindow.

Transactions and blocks are not irreversible when giving a checkpoint on them.
None

ToWorld:
## Description

Make sure transactions and blocks irreversible by the way of giving a checkpoint periodically.

Fixes #426 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Brief of your solution

Give a checkpoint per N block separately.

## How Has This Been Tested?

Regression Test


utxo/utxo_meta.go
```diff
@@ -187,3 +187,75 @@ func (uv *UtxoVM) UpdateForbiddenContract(param *pb.InvokeRequest, batch kvdb.Ba
...
 }
+
+func (uv *UtxoVM) LoadIrreversibleBlockHeight() (int64, error) {
+	irreversibleBlockHeightBuf, findErr := uv.metaTable.Get([]byte(ledger_pkg.IrreversibleBlockHeightKey))
+	if findErr == nil {
+		utxoMeta := &pb.UtxoMeta{}
+		err := proto.Unmarshal(irreversibleBlockHeightBuf, utxoMeta)
+		return utxoMeta.GetIrreversibleBlockHeight(), err
+	}
+	return int64(0), nil
```
fxsjy: it is safe to return 0 only if findErr=="not found"

ToWorld: > it is safe to return 0 only if findErr=="not found"

Thank you for your suggestion and i will correct all of them.

PR #453, opened 2019-10-28
Get gas rate for different resources from genesis json file

ToWorld:
## Description

Make the parameter for gas rate of different resources be a member of genesis json.

Fixes # (issue)

## Type of change

## Brief of your solution

move from hard coding to genesis json file.

## How Has This Been Tested?

Regression Test


cmd/cli/types.go
```diff
@@ -64,6 +64,13 @@ type InvokeRequest struct {
 	ResouceLimits []ResourceLimit   `json:"resource_limits"`
 }
 
+type GasPrice struct {
+	CpuRate  int64 `json:"cpu_rate"`
```
yucaowang: It's better to use uint64 to ensure no negative value.

ToWorld: In order to align  with  the type having existed already, checking the value so that ensuring  that the parameters should not be negative instead.
Please read the latest pull request and this request has been expired.

ToWorld:
Closing this.

PR #462, opened 2019-11-01
fix #461: Contract permission check doesn't include initiator

### LINKED ISSUE #461: Contract permission check doesn't include initiator
**Brief of the issue**
Contract permission check doesn't include initiator, contract call would be failed since ACL not enough.

**Your environment**
XuperUnion master

**Repo steps**
1. Set ACL to a contact interface, give address A permission to call the interface.
2. Invoke or query the interface with address A in as initiator, keep clear of auth_require.

**Expected behaviours**
Have permission to call the interface

**What actually happened**
Returned ACL_NOT_ENOUGH error

**Additional information**

yucaowang:
## Description

1. Combine initiator and auth require in contract permission check.
2. Add identity query interface in unified_check contract.

Fixes #461 

## Type of change

## Brief of your solution

## How Has This Been Tested?


contractsdk/cpp/reserved/unified_check.cc
```diff
@@ -238,6 +241,24 @@ DEFINE_METHOD(UnifiedCheck, identity_check) {
     ctx->ok("success");
 }
 
+// identity_query return whether address is in identity list
+// keep this method for convenience
+DEFINE_METHOD(UnifiedCheck, identity_query) {
```
ToWorld: Is it necessary to add an extra function of ban_query?

yucaowang: no need, banned_check accepts "contract" argument.

PR #491, opened 2019-11-11
New p2p block broadcast mode to improve performance

### LINKED ISSUE #490: Performance improvement on p2p block broadcast
**Brief of your idea**

Now in XuperChain, a new block is broadcasted in full block data mode. Nodes broadcast block data to all neighbours and the neighbours broadcast it again, duplicate block data cannot be checked unless the node received whole block data, this lead to a significant bandwidth cost. 

So we plan to reduce duplicate block broadcast to save bandwidth and computing resources.

Use interactive block broadcast mode: nodes only broadcast new block id, the receiver node Get Block data only if the block is not exist in its ledger.

Describe what might be worse.

**Additional information**

yucaowang:
## Description

In this PR, we propose a new p2p block broadcast mode to reduce bandwidth and computing resources.

Fixes #490 

## Type of change

## Brief of your solution

The block broadcast strategy in v3.4 and before is send full block message in network, and the receiver nodes received the block and broadcast again after validation passed. If each node have 3 peers in average, the block could be received 3 times in p2p layer and processed in xchain for 2 times. 

Say node N0 produced a block B0, then broadcast to its peers N1/N2/N3; N1 received B0 and validated passed, then broadcast B0 to N0/N1/N2/N3; for N0, it received B0 and unmarshal the proto mesage to block data, then check whether the block exists in ledger, definitely B0 is exist, so ignore the message; then N0 received B0 data from N2/N3, the message id is duplicate with B0, so those messages are ignored in p2p layer, but this consumed bandwidth of two block data.

So we proposed new block broadcast mode, the rough idea is new block message is sent in network with only block id, and the receivers ask the sender for full block data only if the block is not exist in receivers' ledger and the new block id is not in processing.

The block broadcast mode could be:

```
// BlockBroadcaseMode is the mode for broadcast new block
//  * Full_BroadCast_Mode = 0, means send full block data
//  * Interactive_BroadCast_Mode = 1, means send block id and the receiver get block data by itself
//  * Mixed_BroadCast_Mode = 2, means miner use Full_BroadCast_Mode, other nodes use Interactive_BroadCast_Mode
//  1. 一种是完全块广播模式(Full_BroadCast_Mode)，即直接广播原始块给所有相邻节点;
//  2. 一种是问询式块广播模式(Interactive_BroadCast_Mode)，即先广播新块的头部给相邻节点，
//     相邻节点在没有相同块的情况下通过GetBlock主动获取块数据.
//  3. Mixed_BroadCast_Mode是指出块节点将新块用Full_BroadCast_Mode模式广播，其他节点使用Interactive_BroadCast_Mode
BlockBroadcaseMode uint8 `yaml:"blockBroadcaseMode,omitempty"`
```

## How Has This Been Tested?


core/xchaincore.go
```diff
@@ -113,6 +116,10 @@ type XChainCore struct {
 	txidCacheExpiredTime time.Duration
 	enableCompress       bool
 	pruneOption          config.PruneOption
+
+	// cache for duplicate block message
+	msgCache           *common.LRUCache
```
ToWorld: common.LRUCache is more expensive than go-cache.

yucaowang: Here I want to limit cache item by capacity not expire time. Using capacity could keep cache in a reasonable memory size, and default 1000 blocks is enough for most case. Since new block is a low frequency message, so the CPU cost could be ignored.

PR #495, opened 2019-11-11
Bugfix: ignore not found when calling db.Has for badger

ToWorld:
## Description

Fix the process of Creating a contract account for the db of badger.

Fixes # (issue)

## Type of change

## Brief of your solution

Ignore the error of errors.New("Key not found"), it is designed for the db of badger.

## How Has This Been Tested?

Regression Test

### DIFF
--- a/xmodel/xmodel.go
+++ b/xmodel/xmodel.go
@@ -213,7 +213,8 @@ func (s *XModel) GetWithTxStatus(bucket string, key []byte) (*xmodel_pb.Versione
 		return nil, false, err
 	}
 	exists, err := s.ledger.HasTransaction(data.RefTxid)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


xmodel/xmodel.go
```diff
@@ -213,7 +213,8 @@ func (s *XModel) GetWithTxStatus(bucket string, key []byte) (*xmodel_pb.Versione
 		return nil, false, err
 	}
 	exists, err := s.ledger.HasTransaction(data.RefTxid)
-	if err != nil {
+	// badger的Has底层还是调用Get接口，需要过滤掉Not found
+	if err != nil && !kvdb.ErrNotFound(err) {
```
icexin: This should be checked at kv layer, not upper layers.

PR #506, opened 2019-11-13
Bugfix: del streams when closed by peer

### LINKED ISSUE #502: Local unconfirmTx cannot be broadcast.
**Brief of the issue**
Describe here

**Your environment**

**Repo steps**
I generate more Tx in the slave node.
The height and utxoTotal is equal all the time.
I found some nodes have more local unconfirmed Tx in the map.

**Expected behaviours**

**What actually happened**

**Additional information**

ToWorld:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
Cover the case that closed the stream when finding the stream has been closed by peer.

Fixes #502 

## Type of change

## Brief of your solution

Check if stream has been closed.

## How Has This Been Tested?

Regression Test

### DIFF
--- a/p2pv2/stream_pool.go
+++ b/p2pv2/stream_pool.go
@@ -153,9 +153,8 @@ func (sp *StreamPool) sendMessage(ctx context.Context, msg *p2pPb.XuperMessage, 
 		return err
 	}
 	if err := str.SendMessage(ctx, msg); err != nil {
-		if err.Error() == "stream reset" || strings.HasSuffix(err.Error(), "connection reset by peer") {
-			sp.DelStream(str)
-		}
+		// delete the stream when error happens
+		sp.DelStream(str)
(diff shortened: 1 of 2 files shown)


p2pv2/stream_pool.go
```diff
@@ -152,7 +153,7 @@ func (sp *StreamPool) sendMessage(ctx context.Context, msg *p2pPb.XuperMessage,
 		return err
 	}
 	if err := str.SendMessage(ctx, msg); err != nil {
-		if err.Error() == "stream reset" {
+		if err.Error() == "stream reset" || strings.HasSuffix(err.Error(), "connection reset by peer") {
```
icexin: Using error string to check for errors is not an idiomatic way.

fxsjy: Yes!

PR #514, opened 2019-11-14
P2p Bugfix

### LINKED ISSUE #502: Local unconfirmTx cannot be broadcast.

ToWorld:
## Description

Deleting the streams when finding that the streams match with the following cases:
1.stream reset
2.connection reset by peer
3.stream closed
4.stream not valid

Fixes #502 

## Type of change

## Brief of your solution

Check if stream has been closed.

## How Has This Been Tested?

Regression Test


p2pv2/stream_pool.go
```diff
@@ -230,7 +227,9 @@ func (sp *StreamPool) sendMessageWithResponse(ctx context.Context, msg *p2pPb.Xu
 	}
 	res, err := str.SendMessageWithResponse(ctx, msg)
 	if err != nil {
-		sp.DelStream(str)
+		if common.NormalizedKVError(err) == common.ErrP2PError {
```
icexin: NormalizedKVError is not a good name for ErrP2PError.
ErrP2PError and the Normalized function should be in p2pv2 package?

qizheng09:
stream limit config need to remove

p2pv2/stream_pool.go
```diff
@@ -27,7 +27,6 @@ type StreamPool struct {
 	log log.Logger
 	// key: peer id, value: Stream
 	streams        *common.LRUCache
-	streamLength   int32
```
qizheng09: stream limit config need to remove

PR #516, opened 2019-11-15
format to print block QC proposal id

qizheng09:
## Description

Format print QC proposal id

## Type of change


cmd/cli/types.go
```diff
@@ -101,7 +101,7 @@ type QCSignInfos struct {
 // QuorumCert is a data type that combines a collection of signatures from replicas.
 type QuorumCert struct {
 	// The id of Proposal this QC certified.
-	ProposalId []byte `protobuf:"bytes,1,opt,name=ProposalId,proto3" json:"ProposalId,omitempty"`
+	ProposalId string `protobuf:"bytes,1,opt,name=ProposalId,proto3" json:"ProposalId,omitempty"`
```
icexin: `HexID` type can be used here.

PR #522, opened 2019-11-17
Bugfix: check if old param and new param's type are both valid

ToWorld:
## Description

Fix bug that only one parameter has been checked and it's not ok.

Fixes # (issue)

## Type of change

## Brief of your solution

Regression Test

## How Has This Been Tested?

### DIFF
--- a/contract/kernel/kernel.go
+++ b/contract/kernel/kernel.go
@@ -290,7 +290,7 @@ func (k *Kernel) validateUpdateIrreversibleSlideWindow(desc *contract.TxDesc) er
 		}
 		switch tp := desc.Args[argName].(type) {
 		case float64:
-			return nil
+			continue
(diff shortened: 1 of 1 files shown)


contract/kernel/kernel.go
```diff
@@ -290,7 +290,7 @@ func (k *Kernel) validateUpdateIrreversibleSlideWindow(desc *contract.TxDesc) er
 		}
 		switch tp := desc.Args[argName].(type) {
```
icexin: You can use the following form to verify the parameters.
``` go
if _, ok := desc.Args[argName].(float64); !ok {
    return fmt.Errorf("......")
}
```

ToWorld: I get it.

PR #570, opened 2019-12-10
contract: add contract call

icexin:
## Description

Different contracts can call each other through the call interface.
The permission check has not been added. @qizheng09 

## Brief of your solution

Add ContractCall interface to Syscall service.

### DIFF
--- a/contract/bridge/syscall_service.go
+++ b/contract/bridge/syscall_service.go
@@ -10,6 +10,7 @@ import (
 	"math/big"
 	"sort"
 
+	"github.com/xuperchain/xuperunion/contract"
 	pb "github.com/xuperchain/xuperunion/contractsdk/go/pb"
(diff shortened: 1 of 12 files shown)


contract/bridge/syscall_service.go
```diff
@@ -130,8 +138,62 @@ func (c *SyscallService) Transfer(ctx context.Context, in *pb.TransferRequest) (
...
+		Core:           nctx.Core,
+		ResourceLimits: *limits,
+		ContractSet:    nctx.ContractSet,
+	}
+	vctx, err := vm.NewContext(cfg)
+	if err != nil {
+		return nil, err
+	}
+	defer vctx.Release()
+	defer func() { delete(nctx.ContractSet, in.GetContract()) }()
```
yucaowang: two consecutive defers could be merged?

icexin: done

PR #575, opened 2019-12-11
Merge small utxos into a bigger one

### LINKED ISSUE #566: Support merging small utxos into a bigger one
**Brief of your idea**
Select small utxo items as many as possible and merge them into one.

Solve the issue that Many small utxo items cause higher query latency, bigger transaction, bigger block.

None

ToWorld:
UTXO Model has one issue that user could have his/her utxo consisting of many small uxtos pieces,
which caused a lower query performance and more fee if the fee model is supported.

In order to solve the problem, the solution that launching a transaction where selecting
utxos as many as possible depending on the size and transfering all selected utxos to yourself
has been proposed.

The type of Utxo includes open utxo, locked utxo and frozen utxo. Where, open utxo means that
the utxo is available and locked utxo means the utxo has been occupied and frozen utxo means that
the utxo couldn't be spent until the specified block height.

You can set the value of open utxo you expected, but you can not merge the rest utxos(locked utxo
and frozen utxo).

The command of `./xchain-cli account merge -N 3` means that you expect to merge all open utxos intoat most 3 uxto pieces.

Fixes #566

## Type of change
## How Has This Been Tested?
Regression Test


cmd/cli/comm_trans.go
```diff
@@ -613,3 +613,54 @@ func printTx(tx *pb.Transaction) error {
...
+		fromAddr, err = readAddress(c.Keys)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
+	utxoInput := &pb.UtxoInput{
+		Bcname:   c.ChainName,
+		Address:  fromAddr,
+		NeedLock: false,
```
yucaowang: Do not need lock UTXO? any change makes utxo conflict in a race condition?

pb/xchain.proto
```diff
@@ -57,6 +57,14 @@ enum TransactionStatus {
 // Xchain is the main interfaces
 service Xchain {
 
+  // MergeUTXO merge many utxos into a few of utxos
+  rpc MergeUTXO(UtxoInput) returns (UtxoOutput) {
```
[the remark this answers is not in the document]
ToWorld: I will fix it. tks

cmd/cli/comm_trans.go
ToWorld: Thank you for your reminder.

PR #579, opened 2019-12-13
Fix the problem of compatibility issues

### LINKED ISSUE #580: Set a default value for GasPrice
**Brief of the issue**
Because of the existence of compatibility issues of GasPrice

ToWorld:
## Description

In order to address compatibility issues of GasPrice.

Fixes #580 

## Type of change

## Brief of your solution

Use a default GasPrice value when the GasPrice configuration is missing
or GasPrice equals 0.

## How Has This Been Tested?

Regression Test

### DIFF
--- a/utxo/utxo_meta.go
+++ b/utxo/utxo_meta.go
@@ -363,6 +363,16 @@ func (uv *UtxoVM) LoadGasPrice() (*pb.GasPrice, error) {
 		if cpuRate < 0 || memRate < 0 || diskRate < 0 || xfeeRate < 0 {
 			return nil, ErrProposalParamsIsNegativeNumber
 		}
+		// To be compatible with the old version v3.3
+		// If GasPrice configuration is missing or value euqals 0, support a default value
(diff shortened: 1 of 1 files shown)


utxo/utxo_meta.go
```diff
@@ -363,6 +363,16 @@ func (uv *UtxoVM) LoadGasPrice() (*pb.GasPrice, error) {
 		if cpuRate < 0 || memRate < 0 || diskRate < 0 || xfeeRate < 0 {
 			return nil, ErrProposalParamsIsNegativeNumber
 		}
+		// To be compatible with the old version v3.3
+		// If GasPrice configuration is missing or value euqals 0, support a default value
+		if cpuRate == 0 && memRate == 0 && diskRate == 0 && xfeeRate == 0 {
```
yucaowang: What if the user DO NOT want to use gas? Can I manually set all rate to 0 in genesis config?

yucaowang: Agreed to solve this issue later in non-gas improvement

PR #586, opened 2019-12-20
Event Subscribe Proto

ToWorld:
Background:
The client can get on-chain data(such as block and transaction) by subscribing a specific event, a kind of interaction between the client and the server(xchain), which is convenient to the client.
Moreover, we can combine it with the asynchronous sdk to improve TPS.

Four events are supported, as follows:
1. BlockEvent
2. TransactionEvent
3. AccountEvent
4. CancelEvent


pb/event.proto
```diff
@@ -0,0 +1,147 @@
...
+  UNSUBSCRIBE_EVENT = 2;
+  BLOCK_EVENT = 3;
+  TRANSACTION_EVENT = 4;
+  ACCOUNT_EVENT = 5;
+}
+
+// BlockStatusInfo 区块元数据
+message BlockStatusInfo {
+  string bc_name = 1;
+  enum EBlockStatus {
```
yucaowang: EBlockStatus is also defined as a internal enum in xchain.proto, can we make it public so we can reuse it?

pb/event.proto
```diff
@@ -0,0 +1,147 @@
...
+    AccountEventRequest account_event_response = 3;
+    SubscribeResponse subscribe_response = 4;
+  }
+}
+
+///// 状态信息
+enum EventType {
+  DEFAULT_EVENT = 0;
+  SUBSCRIBE_EVENT = 1;
+  UNSUBSCRIBE_EVENT = 2;
```
yucaowang: Sub/unsub are event types? what's the meaning of default_event?

pb/event.proto
```diff
@@ -0,0 +1,147 @@
...
+    FURCATION = 2;
+    FAILED = 3;
+    UNCONFIRM = 4;
+  }
+  ETransactionStatus status = 3;
+}
+
+message AccountStatusInfo {
+  string bc_name = 1;
+  enum ETransactionStatus {
```
yucaowang: ditto

pb/event.proto
ToWorld: The status is prepared for BlockStatusInfo only.

pb/event.proto
```diff
@@ -0,0 +1,133 @@
...
+
+package pb;
+
+import "google/api/annotations.proto";
+import "xchain.proto";
+
+service PubsubService {
+  rpc Publish (EventMessage) returns (EventMessage);
+  rpc Subscribe (EventRequest) returns (stream EventResponse);
+  rpc UnSubscribe (EventRequest) returns (stream EventResponse);
```
ToWorld: UnSubscribe doesn't need parameter of stream

pb/event.proto
```diff
@@ -0,0 +1,133 @@
+syntax = "proto3";
+
+package pb;
+
+import "google/api/annotations.proto";
+import "xchain.proto";
+
+service PubsubService {
+  rpc Publish (EventMessage) returns (EventMessage);
```
ToWorld: 内部方法，不需要暴露出来。

PR #593, opened 2019-12-26
Add basic PoA consensus protocol into Xuperchain

### LINKED ISSUE #592: PoA Consensus Implementation
**Brief of your idea**
The PoA protocol is designed for the consortium scene.
The PoA protocol focus on the management of proposers for the Chained-BFT protocol, and provides the liveness for the consensus.

The main parts of PoA are as follows:
- Proposer initialization and update: The proposers in the PoA are initialized in the configuration, and then updated according to the ACL of the specific contract account, which name is also set in the configuration.
- Primary alternation: Like the alternation in TDPoS, every node in the proposer set creates a certain number of blocks successively in a term. And at the start of each term, all nodes will update the proposer set with confirmed update of the ACL.
- Liveness proof: When a node crashes or becomes a byzantine node, the normal nodes will vote for the next proposer. This can continue until a normal node becomes the primary.

It can be applied in more consortium scenes, and provide easy control for CA without voting in TDPoS.
Also, the liveness proof in the PoA can provide a different way from TDPoS to promise the liveness in the consensus, which can be helpful for future consensus protocols.

The communication in the network will be more complicated, because 2 kinds of messages will be added to prove the liveness of PoA consensus.

DhunterAO:
## Description

PoA consensus protocol is necessary in the consortium field. I build the basic PoA implementation which supports the initialization and update of validation set.

 #592 (issue)

## Type of change

## Brief of your solution

The PoA consensus protocol is supported by Chained-BFT and focuses on the primary selection. The validation set is stored in the access control list of the account.

The whole time is divided into different terms. In each term, every validation node in the set will propose new blocks successively. The validation set for the first term is stored in the configuration file(data/config/xuper.json). At the start of a new term(except the first one), the node will find the confirmed ACL of the account and update the validation set accordingly.

## How Has This Been Tested?

The initialization and running of PoA protocol has been tested, including single validation node and multiple validation nodes.
The update of proposers using the ACL of account has been tested.
The liveness of PoA has NOT been tested.


data/config/xuper.json
```diff
@@ -21,25 +21,19 @@
     },
     "new_account_resource_amount": 1000,
     "genesis_consensus": {
-        "name": "tdpos",
```
qizheng09: Add a new chain config named `xuper_poa.json` rather than change this file.

pb/kv_const.go
```diff
@@ -13,6 +13,7 @@ const (
 	VoteProposalPrefix       = "V"
 	PlugConsPrefix           = "P"
 	ConsTDposPrefix          = "D"
+	ConsPoaPrefix            = "A"
```
qizheng09: This prefix seems not be used in this pr.

PR #603, opened 2020-01-06
fix amount of issue and transfer be negative

qizheng09:
## Description

Now the amount of example `xrc01_e1` of `XRC01` can be negative and will be convert to  the Maximum of `uint64`. 

This pull request used to check whether parameters be negative.

## Type of change

### DIFF
--- a/contractsdk/cpp/XRC/xrc01/src/main/xrc01_e1.cc
+++ b/contractsdk/cpp/XRC/xrc01/src/main/xrc01_e1.cc
@@ -18,13 +18,16 @@ bool safe_stoull(const std::string* in, uint64_t* out) {
     if (!in) {
         return false;
     }
-    if (in->substr(0,1) == "-") {
+    size_t first = in->find_first_not_of(' ');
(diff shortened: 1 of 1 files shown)


contractsdk/cpp/XRC/xrc01/src/main/xrc01_e1.cc
```diff
@@ -18,6 +18,10 @@ bool safe_stoull(const std::string* in, uint64_t* out) {
     if (!in) {
         return false;
     }
+    if (in->substr(0,1) == "-") {
```
yucaowang: It's not best way to check a number is signed or not. I cannot see any trim operation on input string, so " -100" could pass this check but result in 18446744073709551516 using stoull.

qizheng09: This function will return `false` and the method will return.

qizheng09: Thanks, this problem has been fixed.

PR #607, opened 2020-01-07
bugfix: unlock useless key when merging utxo

ToWorld:
## Description

When the capacity is over, the key has been locked should be unlocked again.

Fixes # (issue)

## Type of change

## Brief of your solution

When the capacity is over, the key has been locked should be unlocked again.

## How Has This Been Tested?

Regression Test


utxo/merge_utxo.go
```diff
@@ -86,6 +86,7 @@ func (uv *UtxoVM) SelectUtxosBySize(fromAddr string, fromPubKey string, needLock
 		if txInputSize.Cmp(maxTxSize) == 1 {
 			txInputs = txInputs[:len(txInputs)-1]
 			amount.Sub(amount, utxoItem.Amount)
+			uv.unlockKey(key)
```
yucaowang: what if needlock == false?

ToWorld: It's not necessary but better and I have fixed it.

PR #643, opened 2020-01-17
bugfix: no utxo support

### LINKED ISSUE #644: bugfix: the same tx with noutxo model could exist in unconfirmed and confirmed table
**Brief of the issue**
The same tx with noutxo model could exist in unconfirmed and confirmed table, which is not reasonable.

**What actually happened**
image

ToWorld:
## Description

In current case, the tx in the model of no utxo can exist in the unconfirmed and confirmed table both, which caused the native ledger doesn't grow any more.

Fixes #644

## Type of change

## Brief of your solution

take the case where input equals output and input equals 0 into consideration
No need to add a switch for no utxo model

## How Has This Been Tested?

Regression Test

### DIFF
--- a/core/utxo/utxo.go
+++ b/core/utxo/utxo.go
@@ -2439,7 +2439,7 @@ func ValidTxInput(tx *pb.Transaction) error {
 	if tx.GetTxInputs() == nil {
 		return nil
 	}
-	for _, v := range tx {
+	for _, v := range tx.GetTxInputs() {
 		if v != nil {
 			return errors.New("has utxo input")
 		}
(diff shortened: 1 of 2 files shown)


core/utxo/utxo.go
```diff
@@ -1275,6 +1275,15 @@ func (uv *UtxoVM) doTxSync(tx *pb.Transaction) error {
 		uv.xlog.Debug("this tx already in unconfirm table, when DoTx", "txid", fmt.Sprintf("%x", tx.Txid))
 		return ErrAlreadyInUnconfirmed
 	}
+	if TxInputsWithoutUtxo(tx) {
+		exist, err := uv.ledger.HasTransaction(tx.GetTxid())
```
fxsjy: can we release lock before call `HasTransaction`?

PR #650, opened 2020-01-19
Add contract template about award management

ToWorld:
## Description

Add contract template about award management

Fixes # (issue)

## Type of change

## Brief of your solution
use the technology of pure virtual basic class to unify interface
use the erc20 as an implementation 

## How Has This Been Tested?

Regression Test


core/contractsdk/cpp/example/award_manage.cc
```diff
@@ -0,0 +1,388 @@
...
+            ctx->error("missing to");
+            return;
+        }
+
+        const std::string& token_str = ctx->arg("token");
+        if (token_str.empty()) {
+            ctx->error("missing token");
+            return;
+        }
+        int token = atoi(token_str.c_str());
```
yucaowang: Could token be negative or zero?

PR #652, opened 2020-01-20
Add File Hash Contract Template

ToWorld:
## Description

Add File Hash Contract Template
Fixes # (issue)

## Type of change

## Brief of your solution

File Informations, as follows:
1.user_id
2.hash_id
3.file_name

## How Has This Been Tested?

Regression Test


core/contractsdk/cpp/example/hash_deposit.cc
```diff
@@ -0,0 +1,101 @@
...
+    }
+    void storeFileInfo() {
+        xchain::Context* ctx = this->context();
+        std::string user_id = ctx->arg("user_id");
+        std::string hash_id = ctx->arg("hash_id");
+        std::string file_name = ctx->arg("file_name");
+        const std::string userKey = UserBucket + "/" + user_id + "/" + hash_id;
+        const std::string hashKey = HashBucket + "/" + hash_id;
+        std::string value = user_id + "\t" + hash_id + "\t" + file_name;
+        if (ctx->put_object(userKey, value) && ctx->put_object(hashKey, value)) {
```
yucaowang: What if the hashKey already exists? Means this file was stored by other before

core/contractsdk/cpp/example/hash_deposit.cc
```diff
@@ -0,0 +1,101 @@
...
+            ctx->ok("storeFileInfo success");
+            return;
+        }
+        ctx->error("storeFileInfo failed");
+    }
+    
+    void queryUserList() {
+        xchain::Context* ctx = this->context();
+        const std::string key = UserBucket + "/";
+        std::unique_ptr<xchain::Iterator> iter = ctx->new_iterator(key, key + "~");
```
yucaowang: Out of curious, why use 'key + "~"' in this iterator?

ToWorld: ～ as an ending.

PR #659, opened 2020-01-21
add a hashed_timelock example

fxsjy:
## Description
Add a HashedTimelock example for Xuperchain

## Type of change
- [ ] smart contract example

## Brief of your solution
Reference: https://github.com/chatch/hashed-timelock-contract-ethereum/blob/master/contracts/HashedTimelock.sol

## How Has This Been Tested?
manual test


core/contractsdk/go/example/htlc/hashed_timelock.go
```diff
@@ -0,0 +1,130 @@
...
+	oldValue, _ := ctx.GetObject(riddleHash)
+	if oldValue != nil {
+		return code.Errors("duplicated riddle hash:" + string(riddleHash))
+	}
+	expiredHeight, ok := ctx.Args()["expired_height"]
+	if !ok {
+		return code.Errors("missing expired_height")
+	}
+	sender := ctx.Initiator()
+	amount, _ := ctx.TransferAmount()
```
yucaowang: Could amount be negative number or zero?

fxsjy: It should be greater than zero. Actually, it is better to be higher than the tipBlock's height. 
And, I have added a check for args to prevent misstakes of users.

fxsjy: yes, it would be better to check the range of amount here. thanks

PR #663, opened 2020-02-05
exit directly after pruning

ToWorld:
## Description

It is convenient to operate xchain

Fixes # (issue)

## Type of change

## Brief of your solution

Exit directly after pruning

## How Has This Been Tested?

regression test


core/core/xchaincore.go
```diff
@@ -787,7 +788,8 @@ func (xc *XChainCore) Miner() int {
 			xc.pruneOption.Switch = false
 			xc.SyncBlocks()
 			// 裁剪账本可能需要时间，做完之后直接返回
-			continue
+			xc.Stop()
+			os.Exit(0)
```
icexin: Send self a signal?

ToWorld: ok.

PR #667, opened 2020-02-07
Pluggable P2P module implementation

### LINKED ISSUE #666: Abstract P2P interface and make P2P as a pluggable module
**Brief of your idea**

Currently, the P2P module is not pluggable, so KAD network is the only choice of XuperChain users. But in some cases, using gossip and static nodes could be better fitting the real scenarios and requirements.

We could abstract the interface of P2P module, and use pluggable implementation to support different P2P protocols. By using pluggable design, we even could support RPC-based network to meet some private/permission blockchain scenarios.

Describe what might be worse.

**Additional information**

yucaowang:
## Description

This PR is aiming to make P2P module pluggable.

We abstract all P2P interfaces and make the default p2pv2 as a P2P plugin.

Fixes #666 

## Type of change

## Brief of your solution

1. P2P Interfaces

Make `P2PServer` and `Filter` as interface, all P2P plugins should implement these interfaces.

2. Data struct

Keep P2P message as protobuf data definition. The P2P message center and subscriber as common message processing mechanism.

3. P2P plugin

Implemented P2PV2 as the first P2P plugin.

## How Has This Been Tested?


core/p2p/p2pv1/conn.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return nil, err
+	}
+	certPool := x509.NewCertPool()
+	ok := certPool.AppendCertsFromPEM(bs)
+	if !ok {
+		return nil, err
+	}
+
+	certificate, err := tls.LoadX509KeyPair(certPath+"/key.pem", certPath+"/private.key")
```
fxsjy: can ignore err here ?

core/p2p/p2pv1/conn.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func NewConn(lg log.Logger, addr string, certPath, serviceName string, maxMsgSize int) *Conn {
+	conn := &Conn{
+		id:     addr,
+		lg:     lg,
+		quitCh: make(chan bool, 1),
+		lock:   &sync.RWMutex{},
+	}
+	conn.NewGrpcClient(maxMsgSize, certPath, serviceName)
```
fxsjy: NewGrpcClient may be failed ?

core/p2p/p2pv1/server.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		panic(err)
+	}
+	certPool := x509.NewCertPool()
+	ok := certPool.AppendCertsFromPEM(bs)
+	if !ok {
+		panic(err)
+	}
+
+	certificate, err := tls.LoadX509KeyPair(certPath+"/key.pem", certPath+"/private.key")
```
fxsjy: err

core/p2p/p2pv1/conn.go
qizheng09: thanks, fixed

core/p2p/p2pv1/server.go
qizheng09: thanks, fixed

core/p2p/p2pv1/conn.go
qizheng09: thanks, fixed

PR #673, opened 2020-02-12
Support GRPC-based P2P plugin

### LINKED ISSUE #672: p2p module support p2pv1 pluggin which RPC-based
**Brief of your idea**

Currently, the p2p module manage route by KAD and  KAD network is the only choice of XuperChain users. But in some cases, using gossip and static nodes could be better fitting the real scenarios and requirements.

We can provide a new p2p plugin which implement by using grpc. So users can easily define network topology to meet some private/permission blockchain scenarios.

Describe what might be worse.

**Additional information**

yucaowang:
## Description

In this PR, we provide a new P2P plugin which implement by using GRPC called "p2pv1". So users can easily define network topology to meet some private/permission blockchain scenarios.

Fixes #672 

## Type of change

## Brief of your solution

Leveraging the pluggable P2P module design, we implemented all interfaces based on GRPC, the P2P connection is using GRPC streaming and keep long term connections.

How to enable p2pv1:
1. Modify the configuration in conf/xchain.yaml

```
# 区块链节点配置
p2p:
  # module is the name of p2p module plugin, value is [p2pv2/p2pv1], default is p2pv2
  module: p2pv1
  port: 47101
  # the file path of CA certification
  certPath: ./data/tls/1
  # the service name of CA certification
  serviceName: saas_test.server.com
  # static peer infos, user can define different static node list for each blockchain
  staticNodes:
    xuper:
      - "127.0.0.1:47102"
```

 2. make sure the plugins/p2p/p2p-p2pv2.*.so is available and the plugin conf in conf/plugins.conf contains this plugin.

3. start xchain node

## How Has This Been Tested?


core/p2p/p2pv1/conn_pool.go
```diff
@@ -1,36 +1,87 @@
...
+	return nil
+}
+
+// Update add conn to connPool
+func (cp *ConnPool) Update(conn *Conn) error {
+	if cp.conns[conn.GetConnID()] != nil {
+		cp.log.Error("Update conn error", "error", ErrNotExist.Error())
+		return ErrNotExist
+	}
+	cp.lock.Lock()
```
fxsjy: seems it will never goto here.

fxsjy: return ErrAlreadyExist?

qizheng09: fixed

PR #677, opened 2020-02-14
add group chain contract

ToWorld:
## Description

Support group ability that only some nodes have access to one chain.
The allowed list is stored in the group chain contract.

Fixes # (issue)

## Type of change

## Brief of your solution

Support necessary API such store/list/scan and etc..

## How Has This Been Tested?

Regression Test


core/contractsdk/cpp/example/group_chain.cc
```diff
@@ -0,0 +1,168 @@
...
+            std::string target = res.first;
+            int offset = key.length() + 1;
+            int length = target.length() - offset;
+            result += target.substr(offset, length) + endingSeparator;
+        }   
+        ctx->ok(result);
+    }
+    void addNode() {
+        xchain::Context* ctx = this->context();
+        std::string key = nodeBucket + ctx->arg("bcname") + endingSeparator + ctx->arg("ip");
```
fxsjy: arg("bcname") should not be "xuper", may check here

ToWorld: corrected.

PR #678, opened 2020-02-15
Support group chain ability

ToWorld:
## Description

Description, as follows:
One Chain, only allowed nodes can access the network
and broadcast tx and block and etc..
Use contract to maintain the white list.

Fixes # (issue)

## Type of change

## Brief of your solution

Use contract to keep whitelist.
The contract as a kernel contract, which is configured in genesis file.
case1:When receive data from peer node:
Check if the peer node is in whitelist
case2:When broadcast data to peer node:
Check if the peer node is in whitelist

## How Has This Been Tested?

Regression Test


core/core/group_chain.go
```diff
@@ -0,0 +1,117 @@
...
+	StreamCache map[string]string
+	// key: bcname value: map[string]bool
+	StreamContractCache map[string]map[string]bool
+	// key: bcname value: bool
+	ChainContractCache map[string]bool
+	Mutex              *sync.Mutex
+}
+
+// CheckGroupChain 判断某条链下的某个节点是否在白名单中
+func (xm *XChainMG) CheckGroupChain(bcname, remotePeerID string) bool {
```
yucaowang: I think 'FindNode' is more clear than 'CheckGroupChain'.

yucaowang: Or 'FindPeer'

core/utxo/group_chain.go
```diff
@@ -0,0 +1,105 @@
...
+		ResourceLimits: contract.MaxLimits,
+		ContractName:   contractName,
+	}
+	vm, err := uv.vmMgr3.GetVM(moduleName)
+	if err != nil {
+		return 400, nil, err
+	}
+	ctx, err := vm.NewContext(contextConfig)
+	if err != nil {
+		return 400, nil, err
```
fxsjy: ctx not release here?

core/core/group_chain.go
```diff
@@ -0,0 +1,117 @@
...
+	xm.groupChainCache.Mutex.Lock()
+	defer xm.groupChainCache.Mutex.Unlock()
+	xm.groupChainCache.StreamCache = data
+}
+
+func (xm *XChainMG) updateGroupChainCache() {
+	for {
+		xm.updateContractCache()
+		xm.updateStreamCache()
+		time.Sleep(2 * time.Second)
```
fxsjy: magic number, better to be a const or flag

core/p2p/p2pv2/server.go
```diff
@@ -126,10 +138,23 @@ func (p *P2PServerV2) SendMessageWithResponse(ctx context.Context, msg *p2pPb.Xu
 	msgOpts := p2p_base.GetMessageOption(opts)
 	filter := p.getFilter(msgOpts)
 	peers, _ := filter.Filter()
+	fmt.Println("=============>peers:", peers)
```
fxsjy: log is better than print

core/utxo/utxo_meta.go
```diff
@@ -166,6 +166,28 @@ func (uv *UtxoVM) GetForbiddenContract() *pb.InvokeRequest {
 	return uv.meta.GetForbiddenContract()
 }
 
+func (uv *UtxoVM) GetGroupChainContract() *pb.InvokeRequest {
```
yucaowang: I don't suggest to provide functions for different data in UTXO, it's too heavy. I think we can expose a get/set interface in UTXO and passing the key/value of the data to make UTXO more clear.

core/utxo/group_chain.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	res := strings.Split(string(target), "\x01")
+	for _, v := range res {
+		ipMap[v] = true
+	}
+
+	return ipMap
+}
+
+func (uv *UtxoVM) getChainInList() map[string]bool {
```
yucaowang: I think chain node group is a xchaincore feature more than UTXO.

core/p2p/base/option.go
```diff
@@ -7,12 +7,19 @@ type MsgOptions struct {
...
 	// compress
-	Compress bool
+	Compress  bool
+	WhiteList map[string]bool
 }
 
 // MessageOption define single option function
 type MessageOption func(*MsgOptions)
 
+func WithWhiteList(whiteList map[string]bool) MessageOption {
```
yucaowang: Why not use WithTargetPeerAddrs or WithTargetPeerIDs?

core/p2p/p2pv2/server.go
```diff
@@ -105,6 +105,18 @@ func (p *P2PServerV2) SendMessage(ctx context.Context, msg *p2pPb.XuperMessage,
 	msgOpts := p2p_base.GetMessageOption(opts)
 	filter := p.getFilter(msgOpts)
 	peers, _ := filter.Filter()
+	// 还要做一层过滤(使用白名单过滤)
+	peersRes := []peer.ID{}
```
yucaowang: move this logic into getFilter to avoid duplicate code

core/utxo/utxo_meta.go
ToWorld: It is same with ForbiddenContract.

core/utxo/group_chain.go
ToWorld: Application state is related to utxo, so i support an api based on utxo.

core/p2p/base/option.go
ToWorld: Different options, different effects.

core/core/group_chain.go
ToWorld: Corrected with a new name of IsPeerInGroupChain

core/core/group_chain.go
ToWorld: Corrected with a readable name of groupChainCacheUpdateWindow

core/p2p/p2pv2/server.go
ToWorld: I have ignored the log print.

PR #680, opened 2020-02-15
add source_trace contract demo

qizheng09:
## Description

This PR is a cpp contract demo of source trace, the demo contract is aiming to show how to use xuperchain  to build a source trace contract.

This demo have the following functions:

1. Support create a goods by administrator.
2. Support update goods info by administrator and the update event will be recorded by system.
3. Support query the records of  goods by good ID.

## Type of change

### DIFF
--- a/core/contractsdk/cpp/example/source_trace.cc
+++ b/core/contractsdk/cpp/example/source_trace.cc
@@ -1,0 +1,194 @@
+#include "xchain/xchain.h"
+
+
+// 商品溯源合约模板
+// 参数由xchain::Contract中的context提供
+class SourceTrace {
+public:
+    /*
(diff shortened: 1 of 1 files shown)


core/contractsdk/cpp/example/source_trace.cc
```diff
@@ -0,0 +1,194 @@
...
+
+    void initialize() {
+        xchain::Context* ctx = this->context();
+        const std::string& admin = ctx->arg("admin");
+        if (admin.empty()) {
+            ctx->error("missing admin address");
+            return;
+        }
+
+        ctx->put_object("admin", admin);
```
ToWorld: It's better to check the return value for put_object

PR #718, opened 2020-03-06
p2pv1 support nodes join dynamically

qizheng09:
## Description

The p2pv1  doesn't support nodes joined dynamically. It's better to support it.

## Type of change

## Brief of your solution

Support node join and leave dynamically. And the default broadcast contain `staticnodes`, `bootnodes` and `dynamicNodes`.

## How Has This Been Tested?

`p2p` config add `bootNodes`  config as following:
``` 
 bootNodes:
   - "127.0.0.1:47102"
```


core/p2p/p2pv1/server.go
```diff
@@ -108,8 +114,26 @@ func (p *P2PServerV1) Init(cfg config.P2PConfig, lg log.Logger, extra map[string
 			p.staticNodes["xuper"] = peerids
 		}
 	}
-
+	p.connectToBootNodes()
```
yucaowang: user could set both static nodes and bootstrap nodes?

core/p2p/p2pv1/server.go
```diff
@@ -108,8 +114,26 @@ func (p *P2PServerV1) Init(cfg config.P2PConfig, lg log.Logger, extra map[string
...
+
+// ConnectToBootNodes connect to bootnode
+func (p *P2PServerV1) connectToBootNodes() error {
+	msg, err := p2p_base.NewXuperMessage(p2p_base.XuperMsgVersion3, "", "", p2pPb.XuperMessage_NEW_NODE, []byte{}, p2pPb.XuperMessage_NONE)
+	if err != nil {
+		p.log.Error("ConnectToBootNodes NewXuperMessage error", "error", err)
+		return err
+	}
+	msg.Header.From = strconv.Itoa(int(p.config.Port))
+	p.ConnectToPeersByAddr(p.bootNodes)
```
yucaowang: Looks like no peer exchange message?  How can this node get other nodes from bootstrap node?

core/p2p/p2pv1/server.go
qizheng09: User doesn't need to set both `staticNodes` and `bootstrapNodes`.

core/p2p/p2pv1/server.go
qizheng09: A file `msg_processor.go` was missed in last commit. Now p2pv1 is mainly used static nodes as routes. But when new node joins, one nodes needs to restart to update the config. So this commit support nodes joined with not restart any nodes of the network. 
While a nodes joined, bootNodes will add the nodes to it `dynamic nodes` and the default broadcast strategy contain `static nodes`, `boot nodes` and `dynamic nodes`.

PR #722, opened 2020-03-10
Block Header Synchronization

ToWorld:
Support the ability, as follows:
1.Anchor Block
2.Forking handle
3.Maintain the latest block state
4.VerifyTx

## Description

For the use of crossChain between xuperchain and xuperchain.

## Type of change

## How Has This Been Tested?

Regression Test


core/contractsdk/cpp/example/xuper_relayer.cc
```diff
@@ -0,0 +1,477 @@
...
+        return;
+    }
+}
+
+DEFINE_METHOD(XuperRelayer, verifyTx) {
+    xchain::Context* ctx = self.context();
+    const std::string blockid = ctx->arg("blockid");
+    const std::string txid = ctx->arg("txid");
+    const std::string proofPathStr = ctx->arg("proofPath");
+    const int txIndex = atoi(ctx->arg("txIndex").c_str());
```
yucaowang: We do not verify the raw transaction data?

core/contractsdk/cpp/example/xuper_relayer.cc
```diff
@@ -0,0 +1,477 @@
...
+bool within3Confirms(xchain::Context* ctx, const std::string& blockid, const std::string tipBlockid) {
+    int i = 0;
+    std::string currBlockid = tipBlockid;
+    while (i < 3) {
+        if (currBlockid == blockid) {
+            return false;
+        }
+        std::string blockHeaderStr;
+        const std::string currBlockHeaderKey = std::string(blockHeaderBucket) + currBlockid;
+        relayer::InternalBlock* blockHeader = new(relayer::InternalBlock);
```
fxsjy: should this pointer be free later ?

core/contractsdk/cpp/example/xuper_relayer.cc
ToWorld: Issue about verify will be added later.

core/contractsdk/cpp/example/xuper_relayer.cc
ToWorld: Yes, I will corrected it.

core/contractsdk/cpp/example/xuper_relayer.cc
```diff
@@ -0,0 +1,478 @@
...
+        }
+        std::string blockHeaderStr;
+        const std::string currBlockHeaderKey = std::string(blockHeaderBucket) + currBlockid;
+        std::unique_ptr<relayer::InternalBlock> blockHeader(new relayer::InternalBlock);
+        if (!ctx->get_object(currBlockHeaderKey, &blockHeaderStr)) {
+            return false;
+        }
+        blockHeader->ParseFromString(blockHeaderStr);
+        const std::string preHashStr = blockHeader->pre_hash();
+        std::string preHash = std::string(64, 'o');
```
0311xuyang: hash value, not (64, '0') ?

core/contractsdk/cpp/example/xuper_relayer.cc
```diff
@@ -0,0 +1,478 @@
...
+    }
+    std::string visualPreHash = std::string(64, 'o');
+    if (!encode(preHashBuf, visualPreHash)) {
+        ctx->error("encode preHash failed");
+        return;
+    }
+    // 判断preBlockcHeader是否存在
+    const std::string preBlockHeaderKey = std::string(blockHeaderBucket) + visualPreHash;
+    std::string preBlockHeaderStr;
+    std::unique_ptr<relayer::InternalBlock> preBlockHeader(new relayer::InternalBlock);
```
0311xuyang: need to parse preBlockHeader and update the preBlockHeader value?

core/contractsdk/cpp/example/xuper_relayer.cc
ToWorld: Whatever, it's a kind of placeholder.

core/contractsdk/cpp/example/xuper_relayer.cc
ToWorld: Yes, it's necessary. We should verify it by contract self.

PR #723, opened 2020-03-10
optimize the performance of the default sync-mode API of xchain

fxsjy:
## Description
The sync-mode API of xchain is much more convenient, but the performance is poorer than async mode.
This pull rerequest try to make the sync-mode API faster, and the following works are inlcuded.

a. in PostTx API,  "mutex.RLock()" will be used instead of "Lock()" to make posttx requests be processed in parallel, and in order to prevent double-spent or version conflict, a fine-grained spinlock table is used. By the help of spinlock, transactions with race condition will not be handled in paralle.

b.  make the repost window of unconfirmed tx larger, so that it get better performance of syncronization of tx broadcasting.

c.  make the trunk switching of ledger faster, by the help of block cache, in the case of fork occurs.

d.  mark txid in dedup cache only when the DoTx return without problem, whichi can prevent a tx be ignored for a certain long time. Because, a tx failed ago may be success after a while.

e.  bugfix of dirty items are  selected  in utxo cache

f.   bugfix of DAG-parallel verification of transactions in block, which may cause tx be rolledback frequently when a node's block height is a little slower then that of miner.

- [ ] Performance improving (non-breaking change which fixes an issue)

## How Has This Been Tested?

CI passed.
Xu,Yang helps me a lot to do many peroformance test to make it reliable.


core/utxo/spin_lock.go
```diff
@@ -0,0 +1,131 @@
...
+func (sp *SpinLock) TryLock(lockKeys []*LockKey) ([]*LockKey, bool) {
+	succLocked := []*LockKey{}
+	for _, k := range lockKeys {
+		if lkType, occupiedByOthers := sp.m.LoadOrStore(k.key, k.lockType); occupiedByOthers {
+			if lkType == SharedLock && k.lockType == SharedLock { //读读共享
+				sp.refCounter.Add(k.key)
+				succLocked = append(succLocked, k)
+				continue
+			} else {
+				return succLocked, false //读写冲突
```
yucaowang: stop try lock other keys if one failed?

core/utxo/utxo.go
```diff
@@ -146,7 +148,7 @@ type UtxoVM struct {
 	vatHandler           *vat.VATHandler  // Verifiable Autogen Tx 生成器
 	balanceCache         *common.LRUCache //余额cache,加速GetBalance查询
 	cacheSize            int              //记录构造utxo时传入的cachesize
-	balanceViewDirty     map[string]bool  //balanceCache 标记dirty: addr->bool
+	balanceViewDirty     map[string]int   //balanceCache 标记dirty: addr->bool
```
yucaowang: code comment should also be updated.

core/utxo/utxo.go
```diff
@@ -729,6 +732,9 @@ func (uv *UtxoVM) SelectUtxos(fromAddr string, fromPubKey string, totalNeed *big
 				uv.xlog.Trace("utxo still frozen, skip it", "key", string(key), "fheight", uItem.FrozenHeight)
 				continue
 			}
+			keyTuple := bytes.Split(key[1:], []byte("_")) // [1:] 是为了剔除表名字前缀
+			refTxid, _ := hex.DecodeString(string(keyTuple[len(keyTuple)-2]))
+			offset, _ := strconv.Atoi(string(keyTuple[len(keyTuple)-1]))
```
yucaowang: could be a util function for duplicate code?

core/utxo/spin_lock.go
fxsjy: yes ,  and the loser thread should call Unlock to release the locks which has been acquired by itself. 
`
spLockKeys := uv.spLock.ExtractLockKeys(tx)
succLockKeys, lockOK := uv.spLock.TryLock(spLockKeys)
defer uv.spLock.Unlock(succLockKeys)

`

core/utxo/spin_lock.go
```diff
@@ -0,0 +1,131 @@
+package utxo
+
+import "sync"
+import "sort"
+import "strconv"
+import "github.com/xuperchain/xuperchain/core/pb"
+
+const (
```
yucaowang: golint format

core/utxo/utxo.go
fxsjy: thanks, updated

core/utxo/utxo.go
fxsjy: yes, already fix

core/utxo/spin_lock.go
fxsjy: golint ok

PR #724, opened 2020-03-10
Transaction support HD encryption and decryption Info

qizheng09:
## Description

The `xuperchain sdk` will support HD encryption and decryption in near feature. So xuperchain need to add `HDInfo` to support this. 

## Type of change

## Brief of your solution

`xchain.pb` add  `HDInfo` as following,

```
message HDInfo {
  // index
  bytes index = 1;
  // original_hash
  bytes original_hash = 2;
}

```
And the version of TxVersion and the BetaTxVersion upgrade to 2.
```
   TxVersionV2               = 2
   BetaTxVersion             = 2

```


core/utxo/utxo.go
```diff
@@ -89,7 +89,8 @@ const (
 	UTXOCacheSize             = 1000
 	OfflineTxChanBuffer       = 100000
 	TxVersion                 = 1
-	BetaTxVersion             = 1
+	TxVersionV2               = 2
```
yucaowang: What's the meaning of TXVersionV2?

qizheng09: It's no use and have been del.

PR #726, opened 2020-03-11
contractsdk/cpp: add builtin hash function

icexin:
Add builtinResolver to add xvm builtin functions.


core/contractsdk/cpp/xchain/builtins.cc
```diff
@@ -0,0 +1,20 @@
+#include "xchain/builtins.h"
+#include <string>
+
+extern "C" void xvm_hash(const char* name, const char* inputptr, int inputlen,
+                         char* outputptr, int outputlen, int hex_encode);
+
+namespace xchain {
+std::string sha256(const std::string& input) {
```
yucaowang: hash/crypto functions could be aggregated in a sub namespace?

core/contractsdk/cpp/xchain/builtins.cc
```diff
@@ -0,0 +1,20 @@
...
+    char out[64];
+    xvm_hash("sha256", &input[0], input.size(), out, sizeof(out), 1);
+    return std::string(out, sizeof(out));
+}
+
+std::string sha256raw(const std::string& input) {
```
yucaowang: Another implementation way could be providing a bytes to hex string utility, and keep hash return bytes.

core/contractsdk/cpp/xchain/builtins.cc
icexin: Makes sense, has been added.

core/contractsdk/cpp/xchain/builtins.cc
icexin: hex_encode and hex_decode have been added

PR #727, opened 2020-03-12
Support query contracts related to an address

yucaowang:
## Description

This PR supports query contracts related to an address.

This means if an address X is part of ACL of an account Y, then all contracts under account Y is related to address X.

So using this interface, user can query all contracts under all accounts related to an address.

## Type of change

## Brief of your solution

Usage: 
1. Using RPC requests, the RPC interface is:

```
  // GetAddressContracts get contracts of accounts contain a specific address
  rpc GetAddressContracts(AddressContractsRequest)
      returns (AddressContractsResponse) {
    option (google.api.http) = {
      post : "/v1/get_address_contracts"
      body : "*"
    };
  }
```

2. Using xchain-cli for quick test:
./xchain-cli account contracts --address dpzuVdosQrF2kmzumhVeFQZa1aYcdgFpN [--verbose true]

## How Has This Been Tested?


core/pb/xchain.proto
```diff
@@ -1022,3 +1034,19 @@ message ContractStatData {
   int64 accountCount = 1;
   int64 contractCount = 2;
 }
+
+// Query address contracts request
+message AddressContractsRequest {
+  Header header = 1;
+  string bcname = 2;
+  string address = 3;
+  bool need_content = 4;
```
qizheng09: What's the content meaning?

yucaowang: Only contract name, or detailed contract info like txid,timestamp,isbanned etc.

PR #743, opened 2020-03-18
Add cross query proto

qizheng09:
## Description

This pull request add cross query proto, including `CrossChainMeta`, `CrossEndorsorInfo`、`CrossQuery`.


core/pb/xchain.proto
```diff
@@ -1050,12 +1050,36 @@ message AddressContractsResponse {
...
+  string address = 1;
+  string pub_key = 2;
+  // ip+port
+  string host = 3;
+  SignatureInfo signatureInfo = 4;
 }
 
-message CrossQueryResponse { ContractResponse response = 1; }
+message CrossQuery {
+  CrossQueryRequest cross_query_request = 1;
```
icexin: Naming `request` and `response` will be more concise.

qizheng09: `Request` and `Response` is on above. This is used to save to `Bucket`.

PR #750, opened 2020-03-19
Add Hierachical Deterministic Encryption

HawkJing:
## Description

Add Hierachical Deterministic Encryption support to Xuperchain

## Type of change

## Brief of your solution

Hierachical Deterministic Encryption func is added in the crypto client.


HawkJing:
some mistake happened...

PR #758, opened 2020-03-20
Support cross query from other chain

qizheng09:
## Description

This  pull request support cross query from other chain. 

## Type of change

## Brief of your solution

While `PreExec` the  node will query contract from other chain by endorsor server. The endorsor will sign for the cross query responce.

While `PostTX` the node will trust the endorsor, and pass the verify after the sign have been verified.

## How Has This Been Tested?

There is a demo.
```
./core/contractsdk/cpp/example/cross_query_demo/main.cc

```


core/xmodel/cross_query_cache.go
```diff
@@ -0,0 +1,269 @@
...
+		if err != nil {
+			cqc.lg.Info("crossQueryFromEndorsor error", "error", err.Error())
+			return nil, err
+		}
+		cqc.crossQueryCaches = append(cqc.crossQueryCaches, queryInfo)
+		return queryInfo.GetResponse().GetResponse(), nil
+	}
+
+	// 验证背书规则、参数有效性、时间戳有效性
+	crossQuery := cqc.crossQueryCaches[cqc.crossQueryIdx]
```
fxsjy: Crackers may build a fake tx to make it panic here.

core/xmodel/cross_query_cache.go
```diff
@@ -0,0 +1,269 @@
...
+// CrossQuery query contract from otherchain
+func (cqc *CrossQueryCache) CrossQuery(
+	crossQueryRequest *pb.CrossQueryRequest,
+	queryMeta *pb.CrossQueryMeta) (*pb.ContractResponse, error) {
+	cqc.lg.Info("Receive CrossQuery", "crossQueryRequest", crossQueryRequest, "queryMeta", queryMeta)
+	if !isQueryMetaValid(queryMeta) {
+		return nil, fmt.Errorf("isQueryParamValid check failed")
+	}
+	// Call endorsor for responce
+	if cqc.isPenetrate {
```
yucaowang: I think xmodel is a data abstract layer, is it better to implement endorser call in xbridge call?

core/xmodel/cross_query_cache.go
```diff
@@ -0,0 +1,269 @@
...
+	resData, err := json.Marshal(queryInfo.GetResponse())
+	if err != nil {
+		cqc.lg.Info("Marshal Response failed", "err", err)
+		return false
+	}
+	cryptoClient, err := crypto_client.CreateCryptoClient(crypto_client.CryptoTypeDefault)
+	data := append(reqData[:], resData[:]...)
+	digest := hash.UsingSha256(data)
+	for idx := range signsValid {
+		pk, err := cryptoClient.GetEcdsaPublicKeyFromJSON([]byte(signsValid[idx].GetPublicKey()))
```
yucaowang: Do we checked if the endorser address and public key are mach anywhere?

core/xmodel/cross_query_cache.go
```diff
@@ -0,0 +1,269 @@
...
+	if lenCh <= 0 {
+		return nil, nil, errors.New("endorsorQueryWithGroup res is nil")
+	}
+
+	i := 0
+	for r := range msgChan {
+		if i == 0 {
+			conRes = r.CrossQueryResponse
+		} else {
+			if !isMsgEqual(conRes, r.CrossQueryResponse) {
```
icexin: Too expensive equal comparison

core/xmodel/cross_query_cache.go
```diff
@@ -0,0 +1,269 @@
...
+	wg.Wait()
+	// 处理所有请求结果
+	signs := []*pb.SignatureInfo{}
+	var conRes *pb.CrossQueryResponse
+	lenCh := len(msgChan)
+	if lenCh <= 0 {
+		return nil, nil, errors.New("endorsorQueryWithGroup res is nil")
+	}
+
+	i := 0
```
icexin: The meaning of this variable is used to count, called `i` is inappropriate

core/xmodel/cross_query_cache.go
```diff
@@ -0,0 +1,269 @@
...
+			break
+		}
+		i++
+	}
+	return conRes, signs, nil
+}
+
+func (cqc *CrossQueryCache) endorsorQuery(req *pb.EndorserRequest, ce *pb.CrossEndorsor) (*queryRes, error) {
+	ctx, _ := context.WithTimeout(context.TODO(), timeOut*time.Second)
+	conn, err := NewEndorsorConn(ce.GetHost())
```
icexin: Using a short connection per request is too expensive.

core/xmodel/message.go
```diff
@@ -85,3 +86,25 @@ func unmsarshalMessages(p []byte, x interface{}) error {
...
+	encodeIncome, err := encodeMsg(reqIncome)
+	if err != nil {
+		return false
+	}
+	return bytes.Equal(encodeHead, encodeIncome)
+}
+
+func encodeMsg(req proto.Message) ([]byte, error) {
+	var buf proto.Buffer
+	buf.SetDeterministic(true)
```
icexin: Maybe `marshalMessages` should also `SetDeterministic`.

core/xmodel/cross_query_cache.go
```diff
@@ -0,0 +1,269 @@
...
+		go func(req *pb.EndorserRequest, ce *pb.CrossEndorsor) {
+			defer wg.Done()
+			res, err := cqc.endorsorQuery(req, ce)
+			if err != nil {
+				return
+			}
+			msgChan <- res
+		}(req, queryMeta.GetEndorsors()[idx])
+	}
+	wg.Wait()
```
icexin: Waiting for all endorser to return may take a long time, especially if `min_endorsor_num` is relatively small, not all the results are required. It is recommended to send the request while verifying the request, and if the number of signatures is sufficient, canel other requests directly.

core/xmodel/cross_query_cache.go
```diff
@@ -0,0 +1,269 @@
...
+		return nil, err
+	}
+	res := &pb.CrossQueryResponse{}
+	err = json.Unmarshal(endorsorRes.GetResponseData(), res)
+	if err != nil {
+		cqc.lg.Info("endorsorQuery Unmarshal error", "err", err)
+		return nil, err
+	}
+	queryRes := &queryRes{
+		res,
```
icexin: Initialization struct without specifying field names may not pass `go vet`

core/contractsdk/cpp/example/cross_query_demo/src/main.cc
```diff
@@ -0,0 +1,16 @@
+#include "xchain/xchain.h"
+
+struct Hello : public xchain::Contract {};
```
icexin: Replace a suitable contract name

core/contractsdk/cpp/example/cross_query_demo/test/hello.test.js
```diff
@@ -0,0 +1,19 @@
...
+            name: "hello",
+            code: "../hello.wasm",
+            lang: "c",
+            init_args: {}
+        })
+    });
+
+    t.Run("invoke", function (tt) {
+        resp = contract.Invoke("hello", {});
+        assert.equal(resp.Body, "hello world");
```
icexin: Tests is not matching with contract.

core/contract/bridge/cross_uri.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// GetQuery return cross chain uri query
+func (ccu *CrossChainURI) GetQuery() url.Values {
+	return ccu.URL.Query()
+}
+
+// CrossChainScheme define the interface of CrossChainScheme
+type CrossChainScheme interface {
+	GetCrossQueryRequest(*CrossChainURI, []*pb.ArgPair, string, []string) (*xchainpb.CrossQueryRequest, error)
```
icexin: It is better to add names to interface parameters.

qizheng09: Thanks, fixed

core/xmodel/cross_query_cache.go
qizheng09: Now `xmodel` is a data model that can distinguish process `preexec` and `posttx`. Like `model cache`, `utxo cache`, `cross query cache`.

core/xmodel/cross_query_cache.go
qizheng09: But we require that all endorsors's response equal. So cancel other requests directly may not good.

core/xmodel/cross_query_cache.go
qizheng09: `crossQueryNaming` contract store the  public key, so I can use  public key rather than address to verify directly.

core/xmodel/cross_query_cache.go
qizheng09: Thanks, fixed.

PR #759, opened 2020-03-24
add default is any one can create chain config

qizheng09:
## Description

Now user who can create side chain need to in whitelist config. This not convenient in use. 

## Type of change

## Brief of your solution

Kernel config  add `disableCreateChainWhiteList` config. If it configed true, any one can create side chain.

## How Has This Been Tested?

Add `disableCreateChainWhiteList` config as following:

```
kernel:
  disableCreateChainWhiteList:  true

```

### DIFF
--- a/core/common/config/config.go
+++ b/core/common/config/config.go
@@ -257,9 +257,9 @@ type NodeConfig struct {
 
 // KernelConfig kernel config
 type KernelConfig struct {
-	MinNewChainAmount string          `yaml:"minNewChainAmount,omitempty"`
-	NewChainWhiteList map[string]bool `yaml:"newChainWhiteList,omitempty"`
-	IsCreateChain     bool            `yaml:"isCreateChain,omitempty"`
+	MinNewChainAmount          string          `yaml:"minNewChainAmount,omitempty"`
+	NewChainWhiteList          map[string]bool `yaml:"newChainWhiteList,omitempty"`
+	EnableCreateChainWhiteList bool            `yaml:"enableCreateChainWhiteList,omitempty"`
 }
 
 // PruneOption ledger prune option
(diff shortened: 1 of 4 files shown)


core/common/config/config.go
```diff
@@ -259,6 +259,7 @@ type NodeConfig struct {
 type KernelConfig struct {
 	MinNewChainAmount string          `yaml:"minNewChainAmount,omitempty"`
 	NewChainWhiteList map[string]bool `yaml:"newChainWhiteList,omitempty"`
+	IsCreateChain     bool            `yaml:"isCreateChain,omitempty"`
```
icexin: `enableCreateChainWhiteList`  maybe a more clear name. or restructure `NewChainWhiteList` to following?

``` go
type KernelConfig struct {
  ....
  NewChainWhiteList struct {
      Enable bool
      Addrs map[string]bool
  }
  ....
}
```

qizheng09: Thanks, useing `enableCreateChainWhiteList ` here.  Not use `NewChainWhiteList ` struct for backward compatibility.

core/contract/kernel/kernel.go
```diff
@@ -497,8 +503,8 @@ func (k *Kernel) Run(desc *contract.TxDesc) error {
 			return ErrPermissionDenied
 		}
 
-		if !desc.Tx.FromAddrInList(k.newChainWhiteList) {
-			k.log.Warn("tx from addr not in whitelist to create blockchain")
+		if !desc.Tx.FromAddrInList(k.newChainWhiteList) && !k.disableCreateChainWhiteList {
```
yucaowang: Check setting before iterating whitelist could improve performance.

PR #770, opened 2020-03-31
Fix ballot=0 after candidate was revoked

qizheng09:
## Description

While a candidate was revoked, the ballots will be clear. 

## Type of change


core/consensus/tdpos/common.go
```diff
@@ -399,29 +399,47 @@ func (tp *TDpos) validateNominateCandidate(desc *contract.TxDesc) (*cons_base.Ca
...
 	}
-	if len(descNom.Tx.TxInputs) <= 0 {
-		return "", "", "", errors.New("when validateRevokeCandidate, TxInputs should not be nil")
+	canInfo.Address = candidate
+	// process candidate peerid
+	if desc.Args["neturl"] == nil {
+		tp.log.Warn("validateRevokeCandidate candidate have no neturl info",
+			"address", canInfo.Address)
+		// neturl could not be empty when core peers' connection is enabled
+		if tp.config.needNetURL {
```
yucaowang, line 431: I don't think revoke candidate need check neturl, it's only needed when nominating.

core/consensus/tdpos/rollback.go
```diff
@@ -20,31 +21,26 @@ func (tp *TDpos) rollbackVote(desc *contract.TxDesc, block *pb.InternalBlock) er
...
-			canBal := val.(*candidateBallotsCacheValue)
-			if !canBal.isDel {
-				canBal.ballots -= voteInfo.ballots
-				tp.candidateBallotsCache.Store(keyCanBal, canBal)
-			} else {
-				tp.log.Warn("rollbackVote error", "error", "the candidate was revoked!")
-				return nil
-			}
+			canBal := val.(*candidateBallotsValue)
+			canBal.ballots -= voteInfo.ballots
```
yucaowang, line 30: do we need to ensure the votes cannot be negative?

PR #781, opened 2020-04-07
fix verify merkletree when receive block #776

### LINKED ISSUE #776: verify merkletree when receive block
**Brief of the issue**
At present, there is no verification the block's merkletree when receiving blocks. just verification block ID

**Repo steps**
xuperchain source code.
CheckMinerMatch 

**Expected behaviours**
Check whether the block's merkletree is correct

gaojunjie03:
## Description
Fixes #776

## Type of change

## Brief of your solution
add verify merkle root method 

Each consensus plugin CheckMinerMatch method calls the VerifyMerkle method of ledger_hash.go to verify merkleroot

### DIFF
--- a/core/ledger/ledger_hash.go
+++ b/core/ledger/ledger_hash.go
@@ -108,7 +108,7 @@ func encodeJustify(buf *bytes.Buffer, block *pb.InternalBlock) error {
 }
 
 // VerifyMerkle
-func VerifyMerkle(block *pb.InternalBlock)error{
+func VerifyMerkle(block *pb.InternalBlock) error {
 	blockid := block.Blockid
 	merkleTree := MakeMerkleTree(block.Transactions)
 	if len(merkleTree) > 0 {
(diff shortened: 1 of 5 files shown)


core/ledger/ledger_hash.go
```diff
@@ -105,6 +107,23 @@ func encodeJustify(buf *bytes.Buffer, block *pb.InternalBlock) error {
...
 
+// VerifyMerkle
+func VerifyMerkle(block *pb.InternalBlock) error {
+	blockid := block.Blockid
+	merkleTree := MakeMerkleTree(block.Transactions)
+	if len(merkleTree) > 0 {
+		merkleRoot := merkleTree[len(merkleTree)-1]
+		blockMerkleRoot := global.F(block.MerkleRoot)
+		makeMerkleRoot := global.F(merkleRoot)
+		if blockMerkleRoot != makeMerkleRoot {
```
icexin, line 118: No need to convert to a string for comparison, directly use `bytes.Equal(merkleRoot, block.MerkleRoot)` for comparison, reducing memory copy.

gaojunjie03: ok

PR #803, opened 2020-04-30
add data authorization contract

hongyanwang:
## Description

add data authorization contract
update trust operators

## How Has This Been Tested?

passed tests on Linux server


core/contractsdk/cpp/example/trustops/src/trust_counter.cc
```diff
@@ -42,79 +42,151 @@ DEFINE_METHOD(Counter, store) {
...
+      return;
+  }
+  if (!ctx->get_object(ctx->arg("l"), &value)) {
+      ctx->error("get left operand error ");
+      return;
+  }
+  left_op.cipher = value;
+  left_op.commitment = ctx->arg("commitment");
+  // get right operand
+  if (ctx->arg("r").empty() || ctx->arg("commitment2").empty()) {
```
yucaowang, line 65: Suggest check params at the beginning of the function to avoid extra costs.

core/contractsdk/cpp/example/trustops/src/trust_counter.cc
```diff
@@ -42,79 +42,151 @@ DEFINE_METHOD(Counter, store) {
...
-  auto ok = to.add(args_map["l"], args_map["r"], args_map["o"]);
-  std::string debug = "done";
-  if (!ok) {
-    debug = "error";
+  // call trust operator to add, return {{"key", enc_result}}
+  std::map<std::string, std::string> result;
+  auto ok = to.add(left_op, right_op, &result);
+  if (ok) {
+    if (result["key"] == "") {
+        ctx->ok("error");
```
yucaowang, line 81: Suggest use ctx->error to return error code and message

core/contractsdk/cpp/src/xchain/trust_operators/trust_operators.h
```diff
@@ -7,18 +7,35 @@ class TrustOperators {
...
   TrustOperators(xchain::Context *, const uint32_t);
 
-  bool add(const std::string &left_value, const std::string &right_value,
-           const std::string &output_key);
-  bool sub(const std::string &left_value, const std::string &right_value,
-           const std::string &output_key);
-  bool mul(const std::string &left_value, const std::string &right_value,
-           const std::string &output_key);
+  struct Operand {
+    std::string cipher;     //ciphertext
```
yucaowang, line 11: could cipher contains invisible chars? If yes, can we use base64 encoded string?

hongyanwang: ciphertext is base64 encoded string

hongyanwang: > could cipher contains invisible chars? If yes, can we use base64 encoded string?

ciphertext is base64 encoded string

core/contractsdk/cpp/example/trustops/src/trust_counter.cc
hongyanwang: > Suggest check params at the beginning of the function to avoid extra costs.

already fixed

core/contractsdk/cpp/example/trustops/src/trust_counter.cc
hongyanwang: > Suggest use ctx->error to return error code and message

already fixed

PR #810, opened 2020-05-09
separate the status cmd info into sub flag

xiangzepro:
## Description

Extract sub info from status cmd by adding sub flag to simplify the info.

## Brief of your solution

Usage:
  xchain-cli status [flags]

Flags:
  -b, --branch   Get branch info
  -h, --help     help for status
  -l, --ledger   Get ledger info
  -p, --peers    Get peers info
  -u, --utxo     Get utxo info

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,4 +19,4 @@ jobs:
         diff -u <(echo -n) <(gofmt -d $(find . -type f -name '*.go' -not -path "./vendor/*")) 
         make && make test
         if command -v docker && docker info; then XDEV_CC_IMAGE=xuper/emcc:1.38.30 make contractsdk; fi
-        cd ./core/test && bash ci_pipeline.sh
+        cd ./core/test && bash -x ci_pipeline.sh
--- a/core/test/ci_pipeline.sh
+++ b/core/test/ci_pipeline.sh
@@ -12,6 +12,44 @@ outputpath=$(cd `dirname $0`; cd ../..; pwd)
 echo "--------> basepath=$basepath"
 echo "--------> outputpath=$outputpath"
 
+function spawn_process()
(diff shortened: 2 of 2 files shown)


core/cmd/cli/status.go
```diff
@@ -58,6 +75,90 @@ func (s *StatusCommand) printXchainStatus(ctx context.Context) error {
 	return nil
 }
 
+func (s *StatusCommand) convertToFlag() pb.ViewOption {
```
yucaowang, line 78: May I use multi flags in one command?

core/core/xchaincore.go
```diff
@@ -1033,26 +1033,30 @@ func (xc *XChainCore) GetBlockChainStatus(in *pb.BCStatus) *pb.BCStatus {
...
+	if viewOption == pb.ViewOption_NONE || viewOption == pb.ViewOption_UTXOINFO {
+		utxoMeta := xc.Utxovm.GetMeta()
+		out.UtxoMeta = utxoMeta
+	}
 	ib, err := xc.Ledger.QueryBlock(meta.TipBlockid)
 	if err != nil {
 		out.Header.Error = HandlerLedgerError(err)
 		return out
 	}
 	out.Block = ib
```
yucaowang, line 1048: Do not return out.Block if option is not None.

core/cmd/cli/status.go
xiangzepro: Nope, only support one flag currently

core/core/xchaincore.go
xiangzepro: Fixed, thanks

core/cmd/cli/status.go
yucaowang: OK

PR #842, opened 2020-06-15
fix(core/crypto/client): address the potential panic due to concurrent r/w to map

sammyne:
## Description

Currently, the implementation of `cryptoClientFactory` would panic due to concurrent read/write to the internal `clients` **MAP**, if the following two lines at [client/crypto_client_mgr.go](https://github.com/xuperchain/xuperchain/blob/master/core/crypto/client/crypto_client_mgr.go#L33-L54) (exactly at 34th and 50th line) are called at the same time

https://github.com/xuperchain/xuperchain/blob/master/core/crypto/client/crypto_client_mgr.go#L33-L54

## Type of change

## Brief of your solution
This PR makes use of the `RWMutex` like this
- Acquire the read lock, check existence and return if the wanted instance exists
- Acquire the write lock
  - Check existence and return if the wanted instance exists (which is done by the 1st acquirer)
  - Otherwise, construct the wanted instance and return it

> The read lock is used mainly for performance, since successful read should be the most frequent case.

## How Has This Been Tested?

The expected catastrophe happens rarely. I **CANNOT** figure out a good way to reproduce the panic for now :( Therefore, please evaluate how much the PR will help. Discard it if it isn't worthy :)


hello2mao:
LGTM

PR #853, opened 2020-06-19
An erc20 implement for Java contract

Tri-stone:
## Description

An erc20 implement for Java contract.

## Type of change

## How Has This Been Tested?

It has been tested.

### DIFF
--- a/core/contractsdk/java/example/erc20/src/main/java/com/baidu/xuper/Erc20.java
+++ b/core/contractsdk/java/example/erc20/src/main/java/com/baidu/xuper/Erc20.java
@@ -273,11 +273,10 @@ public class Erc20 implements Contract
 
     @ContractMethod
     public Response approve(Context ctx) {
-        byte[] fromByte = ctx.args().get("from");
-        if (fromByte == null) {
-            return Response.error("missing from");
+        String from = ctx.caller();
+        if (from.isEmpty()){
+            return Response.error("missing caller");
         }
-        String from = new String(fromByte);
 
         byte[] toByte = ctx.args().get("to");
         if (toByte == null) {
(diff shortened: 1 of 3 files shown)


core/contractsdk/java/example/erc20/src/main/java/com/baidu/xuper/Erc20.java
```diff
@@ -0,0 +1,323 @@
...
+        String toKey = BALANCEPRE + to;
+        byte[] toBalanceByte = ctx.getObject(toKey.getBytes());
+        BigInteger toBalance;
+        if (toBalanceByte == null) {
+            toBalance = BigInteger.valueOf(0);
+        } else {
+            toBalance = new BigInteger(toBalanceByte);
+        }
+
+        BigInteger fromBalanceNow = fromBalance.subtract(tokenAmount);
```
icexin, line 263: The from party also needs to check whether the balance is sufficient

Tri-stone: fixed

core/contractsdk/java/example/erc20/src/main/java/com/baidu/xuper/Erc20.java
```diff
@@ -0,0 +1,318 @@
...
+            return Response.error("missing to");
+        }
+        String to = new String(toByte);
+
+        byte[] tokenByte = ctx.args().get("token");
+        if (tokenByte == null) {
+            return Response.error("missing token");
+        }
+        String tokenAmountStr = new String(tokenByte);
+        BigInteger tokenAmount = new BigInteger(tokenAmountStr);
```
yucaowang, line 173: What if the token is negative?

Tri-stone: It's fixed. Token amount could not be negative now.

PR #864, opened 2020-07-07
fix compatible issue when upgrade from XPoS

yucaowang:
## Description

Issue: after upgrade a blockchain node using XPoS from v3.7 to v3.8, this node could not send vote message to other peers, and the log shows the peer address(neturl) of all validators are empty, so P2P message sent failed. If the network only have 3 validators, since one node cannot give votes, this network will stop produce new block.

## Type of change

## Brief of your solution

Root cause is the json property names of CandidateInfo struct was changed in v3.8, so the solution is to keep CandidateInfo the same as before.

## How Has This Been Tested?

Tested on a 3-nodes network.

### DIFF
--- a/core/consensus/base/types.go
+++ b/core/consensus/base/types.go
@@ -24,9 +24,9 @@ type ConsensusStatus struct {
 // CandidateInfo define the candidate info
 type CandidateInfo struct {
 	// Address of node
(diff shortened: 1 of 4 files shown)


limeng120:
LGTM

PR #875, opened 2020-07-14
Stop group chain by chain-name.

aucusaga:
## Description
Stop a blockchain by name using transfer smartContract. 

## Type of change

## Brief of your solution
Unload the specific blockchain with XChainmg.

## How Has This Been Tested?
Same way as CreateBlockChain does.


core/contract/kernel/kernel.go
```diff
@@ -62,6 +63,42 @@ var (
 	ErrInvalidChainName = errors.New("Invalid Chain name")
 )
 
+type KernelMethodFunc func(desc *contract.TxDesc) error
+
+type kernelFunc map[string]KernelMethodFunc
+
+func (k *Kernel) getRunKernelFunc() kernelFunc {
+	return kernelFunc{
```
icexin, line 71: This map object should not be created every time it is called. By the way, it doesn't need to create the kernelFunc type, just use the map type as the return value.

aucusaga: Done :)

core/contract/kernel/kernel.go
```diff
@@ -495,104 +526,124 @@ func (k *Kernel) validateUpdateBlockChainData(desc *contract.TxDesc) error {
...
+	}
+	if bcName == "xuper" {
+		k.log.Warn("Oops, main-chain:xuper cannot be stopped.")
+		return ErrPermissionDenied
+	}
+	if !desc.Tx.FromAddrInList(k.newChainWhiteList) && !k.disableCreateChainWhiteList {
+		k.log.Warn("Address in tx for stopping a blockchain should be in whitelist", "disableCreateChainWhiteList", k.disableCreateChainWhiteList, "ChainWhiteList", k.newChainWhiteList)
+		return ErrAddrNotInWhiteList
+	}
+	if err := k.register.UnloadBlockChain(bcName); err != nil {
```
chunhui01, line 582: If xchain restarted, it will be reloaded,whether the directory can be renamed?

aucusaga: Moving deleted-chain dictionary is complicated when deal with rollback operation, which should be atomic, so we put it forward.

qizheng09: > If xchain restarted, it will be reloaded,whether the directory can be renamed?

User need to move the chain to other folder after it has been stopped.

PR #876, opened 2020-07-14
*: refactoring event mechanism and add contract event support

icexin:
## Description

refactoring event mechanism and add contract event support.

## Type of change

- [x] This change requires a documentation update

## Brief of your solution

Add a breakpoint mechanism based on block height to restore the subscription position, and unify the filter rules.

## How Has This Been Tested?

unit test

### eventService
The code is in `./core/server/event_service.go`
Grpc service for event service, at the same time as the entrance of traffic to check the number of connections.

### Router
The code is in `./core/event/router.go`
Event distributor, according to the event type and filter to return the corresponding event iterator

### Topic
The code is in `./core/event/topic.go`
This interface defines the interface that an event topic needs to implement.

``` go
// Topic is the factory of event Iterator
type Topic interface {
    // ParseFilter deserializes the topic filter from the specified bytes buffer
    // The returned parameter will be passed as the input parameter to the filter parameter of NewIterator
    ParseFilter(buf []byte) (interface{}, error)

    // MarshalEvent encode event payload returns from Iterator.Data()
    MarshalEvent(x interface{}) ([]byte, error)

    // NewIterator make a new Iterator base on filter
    NewIterator(filter interface{}) (Iterator, error)
}
```

### Iterator
The code is in `./core/event/topic.go`
This interface defines an event stream, provided as an iterator.
``` go
// Iterator is the event iterator, must be closed after use
type Iterator interface {
    Next() bool
    Data() interface{}
    Error() error
    Close()
}
```

### BlockTopic
The code is in `./core/event/block_topic.go`
Implemented the `Topic` interface. For events with a subscription type of `BLOCK`, it returns a corresponding iterator that returns `pb.FilterBlock` according to `pb.BlockFilter`.

### BlockIterator
The code is in `./core/event/block_iterator.go`
Implemented the `Iterator` interface, which returns a block stream of the specified range. If there is no new block, it will block on the `Next` method.

### filteredBlockIterator
Iterator combining `BlockIterator` and `pb.BlockFilter` returns the filtered block.


core/utxo/utxo.go
```diff
@@ -1846,7 +1851,7 @@ func (uv *UtxoVM) Walk(blockid []byte, ledgerPrune bool) error {
 				return updateErr
 			}
 		}
-		updateErr := uv.updateLatestBlockid(undoBlk.PreHash, batch, "error occurs when undo blocks")
+		updateErr := uv.updateLatestBlockid(undoBlk.PreHash, block.Height, batch, "error occurs when undo blocks")
```
chunhui01, line 1854: The height and block don't match here,will cause abnormal.

core/xmodel/message.go
```diff
@@ -13,8 +13,8 @@ var (
 	protoIface = reflect.TypeOf((*proto.Message)(nil)).Elem()
 )
 
-// marshalMessages marshal protobuf message slice
-func marshalMessages(msgs interface{}) ([]byte, error) {
+// MarshalMessages marshal protobuf message slice
+func MarshalMessages(msgs interface{}) ([]byte, error) {
```
chunhui01, line 17: Is it better for this method to belong to the common package?

core/utxo/utxo.go
icexin: fixed

PR #877, opened 2020-07-16
utxo/txhash: new txhash algorithm

icexin:
## Description

Introduce a new txhash algorithm

## How Has This Been Tested?

unit test


core/utxo/txhash/encode.go
```diff
@@ -0,0 +1,183 @@
...
+	case int64:
+		e.EncodeInt64(v)
+	case string:
+		e.EncodeString(v)
+	case []byte:
+		e.EncodeBytes(v)
+	case map[string][]byte:
+		e.EncodeMap(v)
+	default:
+		log.Panicf("not supported type:%T", x)
```
yucaowang, line 86: Why use Panicf here? It's better to throw an exception?

PR #879, opened 2020-07-20
xpoa change: use snapshot of xmodel to get latest valid proposers in …

aucusaga:
## Description
Use snapshot of xmodel and rewrite xpoa_validates.cc.

## Type of change


core/consensus/xpoa/xpoa.go
```diff
@@ -351,8 +351,14 @@ Again:
...
 // 注意：当查不到的时候或者一个候选人都没查到则默认取初始化的值
 // TODO: zq needs to be optimized in future because
-func (xpoa *XPoa) getCurrentValidates() ([]*cons_base.CandidateInfo, int64, int64, error) {
-	contractRes, confirmedTime, confirmedHeight, err := xpoa.utxoVM.SystemCall(xpoa.xpoaConf.contractName, xpoa.xpoaConf.methodName, nil, true)
+func (xpoa *XPoa) getCurrentValidates(curHeight int64) ([]*cons_base.CandidateInfo, int64, int64, error) {
+	preBlockId, err := xpoa.ledger.QueryBlockByHeight(curHeight)
+	xmod := xpoa.utxoVM.GetXModel()
+	reader, err := xmod.CreateSnapshot(preBlockId.GetBlockid())
+	if err != nil {
+		reader = xmod
```
chunhui01, line 359: If set xmod, it may cause a bug. whether the default value should be returned?

core/consensus/xpoa/xpoa.go
```diff
@@ -351,8 +351,14 @@ Again:
 // getCurrentValidates return current validates from xmodel
 // 注意：当查不到的时候或者一个候选人都没查到则默认取初始化的值
 // TODO: zq needs to be optimized in future because
-func (xpoa *XPoa) getCurrentValidates() ([]*cons_base.CandidateInfo, int64, int64, error) {
-	contractRes, confirmedTime, confirmedHeight, err := xpoa.utxoVM.SystemCall(xpoa.xpoaConf.contractName, xpoa.xpoaConf.methodName, nil, true)
+func (xpoa *XPoa) getCurrentValidates(curHeight int64) ([]*cons_base.CandidateInfo, int64, int64, error) {
+	preBlockId, err := xpoa.ledger.QueryBlockByHeight(curHeight)
+	xmod := xpoa.utxoVM.GetXModel()
```
qizheng09, line 356: May be it's better to package this process in SysCall function. Or utxovm return an snapshot reader.

core/utxo/system_call.go
```diff
@@ -42,16 +41,15 @@ func (uv *UtxoVM) SystemCall(contractName, methodName string, args map[string][]
...
 	if invokeErr != nil {
 		ctx.Release()
 		return nil, lastConfirmedTime, lastConfirmedHeight, invokeErr
 	}
 	rset, _, _ := modelCache.GetRWSets()
 	ctx.Release()
-	if !withConfirmed {
-		return invokeRes.Body, lastConfirmedTime, lastConfirmedHeight, nil
-	}
 	for _, v := range rset {
```
qizheng09, line 53: This no more use while use snapshot reader.

core/consensus/xpoa/xpoa.go
aucusaga: Default configuration might cause unavoidable mistake since default proposers might have been deleted while the system continue to use them.

core/utxo/system_call.go
aucusaga: Done :)

core/consensus/xpoa/xpoa.go
aucusaga: SystemCall is a general function with different XMReader, say, XModel or SnapShot, for preparing model cache env to invoke contract. In my opinion, preparing XMReader outside the SystenCall function might a better way in view of further general usage.

aucusaga: Utxovm return an snapshot reader: done :)

core/consensus/xpoa/xpoa.go
aucusaga: Finally, it's seem to be more reasonable while return error with nil value, where users can handle the specific error under no risk.

PR #912, opened 2020-09-10
fix bug of v3.10 xpoa-opti

aucusaga:
v3.10 fix xpos bug: validate the smr state when update consensus from tdpos to tdpos.
v3.10 fix xpoa bug: the number of txs which modify validates should not exceed the threshold 1.


chunhui01 (approved):
LGTM

qizheng09 (approved):
LGTM

PR #936, opened 2020-10-19
Xpoa::Add param blockid in xpoa contract.

aucusaga:
Add a param named blockid in xpoa contract, which limits the concurrent operation of the user.


qizheng09:
Please remove traffic restrictions until the consensus stability plan ends.

PR #938, opened 2020-10-20
Added tx endorser

preminem:
## Description
Bitcoin, Ethereum and other blockchain transactions are packaged in the Merkel tree structure, so the light client only needs to store the block header to verify whether a transaction is on the chain through SPV (Simple Payment Verification).

Verify the validity of the transaction, which is used in many scenarios. For example, Alice wants to purchase a product, and after sending the transaction transfer to the merchant Bob, Bob can verify the validity of the transaction through the light client, and then give the product to Alice.

In addition, the prerequisite for achieving cross-chain transactions is that the target chain can verify the existence of a transaction on the source chain, and the validity verification of the transaction is also required.


core/server/xendorser/plugin-default/xendorser_impl_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"os"
+	"testing"
+	"time"
+
+	"github.com/xuperchain/xuperchain/core/pb"
+)
+
+var baseDir = os.Getenv("XCHAIN_ROOT")
+
+type MorkXEndorserServer struct {
```
icexin, line 15: `s/Mork/Mock`

core/server/xendorser/plugin-default/xendorser_impl_test.go
```diff
@@ -0,0 +1,80 @@
...
+}
+func (s *MorkXEndorserServer) PreExecWithSelectUTXO(context.Context, *pb.PreExecWithSelectUTXORequest) (*pb.PreExecWithSelectUTXOResponse, error) {
+	return nil, nil
+}
+func (s *MorkXEndorserServer) PreExec(context.Context, *pb.InvokeRPCRequest) (*pb.InvokeRPCResponse, error) {
+	return nil, nil
+}
+
+func TestGetTx(t *testing.T) {
+	os.Chdir(baseDir)
```
icexin, line 33: Check `basedir` variable, if it is empty, skip the test

core/server/xendorser/plugin-default/xendorser_impl.go
```diff
@@ -5,21 +5,34 @@ import (
 	"errors"
 	"io/ioutil"
 
-	context "golang.org/x/net/context"
+	"golang.org/x/net/context"
```
icexin, line 8: Switch to the `context` package of the standard library

chunhui01 (approved):
LGTM

core/server/xendorser/plugin-default/xendorser_impl_test.go
preminem: done

core/server/xendorser/plugin-default/xendorser_impl_test.go
preminem: done

core/server/xendorser/plugin-default/xendorser_impl.go
preminem: No change now

icexin (approved):

PR #941, opened 2020-11-03
Fix contract concurrent bug

preminem:
## Description

Fix the concurrency bug when the contract method contains event trigger. When locking the key in the output, if the bucket contains ¥transient, skip it.


icexin (approved):

qizheng09 (approved):
LGTM

PR #961, opened 2020-12-29
LedgerKeeper::fix utxoVM & ledger unequal error.

aucusaga:
PlayAndRepost might fail when the ledger tries to confirm a block, which causes return action in the code while the block has been put into the ledger incorrectly.
Unfortunately, a new sync task will continue to confirm a block firstly, which will cause the ledger's status ahead of utxoVM by 2 block-height and make utxoVM play to fail.

### DIFF
--- a/core/core/ledgerkeeper.go
+++ b/core/core/ledgerkeeper.go
@@ -348,8 +348,12 @@ func (lk *LedgerKeeper) getPeerBlockIds(beginBlockId []byte, length int64, targe
 		lk.log.Warn("ledgerkeeper::getPeerBlockIds::Sync Headers P2P Error: local error or target error", "Logid", msg.GetHeader().GetLogid(), "Error", err)
 		return true, nil, ErrInternal
 	}
-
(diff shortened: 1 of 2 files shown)


core/core/ledgerkeeper.go
```diff
@@ -348,8 +348,12 @@ func (lk *LedgerKeeper) getPeerBlockIds(beginBlockId []byte, length int64, targe
...
-
-	response := res[0]
+	var response *xuper_p2p.XuperMessage
+	for _, msg := range res {
+		if msg.GetHeader().GetBcname() == lk.bcName {
+			response = msg
+		}
+	}
 	headerMsgBody := &pb.GetBlockIdsResponse{}
 	err = proto.Unmarshal(response.GetData().GetMsgInfo(), headerMsgBody)
```
chunhui01, line 358: If the response is nil, it will cause panic.

chunhui01 (approved):

wellzm (approved):

PR #971, opened 2021-01-10
RFC: Add debug output for xdev && Allow options args to specify account  for unit test

chenfengjin:
## Description

- Add Debug output for unit test when  DEBUG environment variable set, you can easily inspect resp.body and resp.message when assert failed without change youre test.js file 

- Allow specify account for unit test

Fixes # (issue)

## Type of change

### DIFF
--- a/core/cmd/xdev/internal/jstest/xchain/environment.go
+++ b/core/cmd/xdev/internal/jstest/xchain/environment.go
@@ -62,10 +62,6 @@ func newEnvironment() (*environment, error) {
 	}, nil
 }
 
-type OptionArgs struct {
-	Account string `json:"account"`
-	Amount  string `json:"amount"`
-}
 type deployArgs struct {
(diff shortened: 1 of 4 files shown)


core/cmd/xdev/internal/jstest/xchain/environment.go
```diff
@@ -61,13 +62,18 @@ func newEnvironment() (*environment, error) {
 	}, nil
 }
 
+type OptionArgs struct {
```
icexin, line 65: 这个地方最好复用invokeOption

icexin (approved):

core/cmd/xdev/internal/jstest/xchain/environment.go
chenfengjin: Done

Tri-stone (approved):
LGTM

PR #1123, opened 2022-02-16
bugfix: stop voting height checking use xuper chain height even when using parachain

chenfengjin:
## Type of change


godeamon (approved):

zheng-bin (approved):
LGTM

